PR #10, opened 2023-04-13
Cat equiv

stschaef:
Established that currying is an equivalence of categories. Some notes about this:

- I renamed everything to make more sense. Maybe a little too verbose, but the public parts seem to have nice enough names. And the non-public parts at least have descriptive names
- Not sure how to organize the `module`s in the code so that only the parts we want are exposed in the public interface
- I define a `swapArgs` functor, so that we can left-hand curry as well. Show that `swapArgs` is an equivalence doesn't seem to play nicely with Agda's termination checker. I'm not doing any recursion by hand, but perhaps `solveCat!` is inducing some strange behavior here
- After talking some with Pranav, it seems like it might be useful to build a proper uncurry functor, as we have all of the data for it laying around in these proofs

These last two points are not complete, but I think there may be enough that is complete to unblock the other issues

### DIFF
--- a/Cubical/Categories/Instances/Functors/More.agda
+++ b/Cubical/Categories/Instances/Functors/More.agda
@@ -121,8 +121,15 @@ module _ (C : Category ℓC ℓC') (D : Category ℓD ℓD') where
         λη .N-ob γ .N-ob c ∎))
       ) ) ∣₁)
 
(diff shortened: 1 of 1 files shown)


maxsnew:
Ok LGTM. I invited you to be a collaborator so if that's working then merge yourself

PR #11, opened 2023-04-16
Comment out left-handed currying

stschaef:
Still getting termination problems with code that seems like it should properly define swap, but I'm merging in the proof of equivalence for the right-handed currying. 

Commenting out the non-working parts involving the swap functor, as there are still holes


maxsnew:
Wait are you saying that the code you just checked in has termination problems?

stschaef:
Im saying that the right-handed currying equiv of cats is 100% good to go. And the left-handed version has a termination issue related to this swap functor, which is why I initially meant to comment it out 

My hope was that even the right-handed definition would be sufficient to unblock the other issues. But think I was a little too sloppy here. 

I will fix that issue up all the way, and try not to spam PRs 🙃

maxsnew:
Can you point me to the termination problem (or continue this convo on slack)? I think I know what the issue is

PR #13, opened 2023-04-24
Removing Univalence Dependency in Curry Equivalence Proof

GenericMonkey:
@stschaef and I ran into some univalence propagation issues in Profunctor because of assuming univalence in earlier proofs. 

Repackaged some proofs to explicitly construct the equivalence, and also got a general proof of equivalences composing that didn't rely on `NatIsotoPath` or vice versa. Removes the univalence requirement, allowing Profunctor to simply be defined using `curryF`

### DIFF
--- a/Cubical/Categories/Equivalence/More.agda
+++ b/Cubical/Categories/Equivalence/More.agda
@@ -18,6 +18,7 @@ open import Cubical.Categories.Equivalence.Base
 open import Cubical.Categories.Category
 
 open import Cubical.Tactics.CategorySolver.Reflection
+open import Cubical.Categories.NaturalTransformation.More
 
 private
   variable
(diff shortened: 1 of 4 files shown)


stschaef (approved):
Looks good to me

maxsnew:
Do we get a stronger property with univalence? Like that these representability notions are Path-equal?

GenericMonkey:
I'm not sure, I don't think I have a good enough grasps on paths to answer that. However, if we do need that stronger notion, I think we can add Univalence as an additional assumption detached from these proofs and get that result in a separated module?

maxsnew:
That's true. The main downside is that it might be *easier* to prove the version where we additionally assume univalence.

But the upstream cubical lib works with univalence as a separate assumption so for this proof it makes sense to follow their style

PR #22, opened 2023-06-01
Rezk by HIT

hejohns:
Exercise to learn to work with higher inductive types.
Following the book proof (ch 9.9, second proof) but currently stuck on showing that the reflexive path on C [ a , b ] from isoToPath, where the isomorphism is by postcomposing by 1_b both ways, is equal to refl C [ a , b ]
-- if it's even true?
Tabling this to work on the (binary) cartesian product solver.
 
Also TODO: copy over category lift from my fork so the interface to the rezk completion doesn't force the user to lift their (pre)category themselves.

### DIFF
--- a/Cubical/Categories/RezkCompletion/Constructionn.agda
+++ b/Cubical/Categories/RezkCompletion/Constructionn.agda
@@ -1,0 +1,93 @@
+{-
+
+The Construction of Rezk Completion
+
+-}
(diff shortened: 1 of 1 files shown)


maxsnew:
Subsumed by https://github.com/maxsnew/multi-poly-cats/pull/53

PR #25, opened 2023-06-04
Bifunctor prof

stschaef:
We now have the 4 definitions of profunctor representability involving bifunctors and the functions going between these definitions.

The proof that these are equivalences is still yet to come, but I want to merge in the functions for usage


stschaef:
I did not mean to include `Equiv23.agda` here, but this file is not imported anywhere. It is just a placeholder until we update `Equivalence.agda`, and because the equivalences are not updated right now anyway it is safe to merge this in

PR #27, opened 2023-06-08
Set Up CI 

stschaef:
This sets up CI with two checks

1. line length is less than 80 chars. In the future we can hopefully swap this out for a real linter, but I haven't found any nice agda linter. So this is a very basic style check. I went through and fixed this on many files, but there is still some left.
2. checks if all code compiles

Note, we will "fail" builds for the immediate future because there exists currently broken code on main, and there is still some code with >80 char width. 

In terms of behavior, the agda code in main is the same as before, so I don't think failed builds in this sense is a problem for the immediate future. Over the next few days we can finish the style fix and moving broken code into new branches


maxsnew:
Do you know about `git rebase`?

maxsnew:
Well don't change the history once it's pushed to main. But in the future please rebase. I'll change the pull request rules to make rebasing the default

PR #35, opened 2023-06-16
Refactor of Poset Cat with Grothendieck Construction

GenericMonkey:
- stole from other repo once again, defining `Preorder`
- Redefined monotone function to be between `Preorders`, defined category of `Preorders`
- Defined `DisplayedCategory`, `DisplayedPoset`, with a map allowing promotion from the latter to the former. Defined the Grothendieck construction for a displayed category
- Reworked Monotone adjoints as `Prop`s. Defined `Left` `Right` and `Both` modularly, giving us all the categories for free


GenericMonkey:
The Beck-Chevally condition got a bit clunky because I used propositional truncation on the adjoint properties. If you have a nice way to remedy this, i'm all ears

maxsnew:
If it's a poset, you don't need truncation because the adjoints are unique if they exist

GenericMonkey:
That makes sense, will rework `∃[ L ∈ MonFun Y X ] (L ⊣ f)` into `Σ[ L ∈ MonFun Y X ] (L ⊣ f)` then. One question that I'm still not sure about is whether or not we need truncation on the `(L ⊣ f)` part. Can there be more than one isomorphism between the 2 props in the adjoint defn?

maxsnew:
No, Iso between two Props is a Prop. Should be able to prove it using some consequences of univalence in the stdlib:
1. Props are Sets
2. Iso between Sets is equivalent to Path
3. Prop is a Set.
Therefore Iso is in this case a Prop

GenericMonkey:
Finished up the proof the full sigma type of adjoint info is a prop. I'm pretty happy with how clean the BC condition is now, as well as how nice defs of new cats are. Lmk if you have any other suggestions/refactor ideas

Cubical/Categories/Constructions/Subcategory.agda
```diff
@@ -0,0 +1,100 @@
...
+  variable
+    ℓC ℓC' ℓD ℓD' ℓE ℓE' ℓP ℓQ ℓR : Level
+
+module _ (C : Category ℓC ℓC') where
+  private
+    module C = Category C
+  open Category
+
+
+  record DisplayedCategory : Type (ℓ-max (ℓ-max ℓC ℓC') (ℓ-suc ℓP)) where
```
maxsnew, line 22: This should go in its own file

maxsnew (changes requested):
Looks good. Should just spin off displayed cat to its own file

Cubical/Categories/Constructions/Subcategory.agda
GenericMonkey: Moved to `DisplayedCategory.agda`. Left the `DisplayedPoset` and `Grothendieck` in the Subcategory file though

Cubical/Categories/Constructions/Subcategory.agda
```diff
@@ -0,0 +1,73 @@
...
+    ; D-⋆IdR = λ k →
+        isProp→PathP (λ i → isPropHomf {f = ((C .⋆IdR _) i)}) _ _
+    ; D-⋆Assoc = λ k l m →
+        isProp→PathP (λ i → isPropHomf {f = ((C .⋆Assoc _ _ _) i)}) _ _
+    } where
+    open DisplayedPoset D
+
+  -- the Grothendieck construction, or the generalized construction
+  -- for a subcategory
+  Grothendieck : DisplayedCategory {_} {_} {ℓP} C → Category _ _
```
maxsnew, line 58: This should go in its own file and construct the following things:
1. The projection functor from Gro C D -> C
2. Show that if the displayed category is a displayed poset then that projection functor is faithful. (In fact its an iff)

GenericMonkey: Just to make sure I understand the separation here:
- `DisplayedCategory` should be in `DisplayedCategory.agda`
- `DisplayedPoset` and `DisplayedPoset->Cat` should be in its own file (`DisplayedPoset.agda`?)
- `Grothendieck` + the functor argument should be in its own file (rename `Subcategory.agda` -> `Grothendieck.agda`)?

maxsnew: For `Grothendieck` it should be in something like `DisplayedCategory.Grothendieck`, since there are other Grothendieck constructions, but otherwise yes

GenericMonkey: Cool. Is making `DisplayedPoset.agda` into `DisplayedCategory.DisplayedPoset.agda` a good idea as the latter builds on the first? Or keep it as its own construction top level?

maxsnew: Sure makes sense. Then go ahead and squash and push to main

PR #39, opened 2023-07-11
update to cubical HEAD

maxsnew:
I fixed most things except:
1. Anything outside the `Cubical` subdirectory is considered obsolete. I changed the Makefile to only check things in `Cubical`.
2. `Profunctor.General` I only implemented one function which I called `FunctorComprehension` in line with the discussion at https://github.com/maxsnew/multi-poly-cats/issues/36 . `Profunctor.Equiv` is not fixed at all.


Cubical/Categories/Bifunctor/Redundant.agda
stschaef, line 1: I'm guessing `Sep` means something like "separate" throughout this file, but what does `Par` mean here?

stschaef (approved):

Cubical/Categories/Bifunctor/Redundant.agda
maxsnew: Parallel

PR #42, opened 2023-07-20
Freecat refactor

maxsnew:
This is a refactor of the UMP of free categories, quotient categories and presented categories to use an eliminator into a displayed category. This makes their eliminators much more like the dependently typed `elim`inators for data types such as `Cubical.Data.Nat.Base.elim`, where we think of displayed categories as the category analogue of a dependent type. This required defining a `Section` of a displayed category, which is the displayed category equivalent of `(x : A) -> P x`.

Also refactored presented categories to use quotient categories, which I didn't realize were already in cubical.

Finally, I also unrelatedly defined displayed category `Slices` and displayed preorder `Subobjects`.


Cubical/Categories/Displayed/Category/More.agda
GenericMonkey, line 1: I didn't see an upstream `Displayed.Category` -- did you want to add this to `Displayed.Base.More`?

Cubical/Categories/Displayed/Section.agda
```diff
@@ -0,0 +1,128 @@
...
+    (G .F-seq f g)
+    i
+
+  SectionToFunctor : Section → Functor C (∫C D)
+  SectionToFunctor F .F-ob x = x , F .F-ob x
+  SectionToFunctor F .F-hom f = f , (F .F-hom f)
+  SectionToFunctor F .F-id = ΣPathP (refl , (F .F-id))
+  SectionToFunctor F .F-seq f g = ΣPathP (refl , (F .F-seq f g))
+
+  Fst : Functor (∫C D) C
```
GenericMonkey, line 60: Think this is already defined in Displayed.Base.More now

Cubical/Categories/Displayed/Category/More.agda
maxsnew: I think the upstream `Displayed.Base` should probably be renamed to `Displayed.Category.Base` and then this file should probably be `Displayed.Category.Properties`.

maxsnew: Oh I see, you already made `Displayed.Base.More`

Cubical/Categories/Displayed/Slice.agda
```diff
@@ -0,0 +1,52 @@
...
+    ℓC ℓC' ℓD ℓD' : Level
+
+open Category
+open Categoryᴰ
+open Preorderᴰ
+module _ (C : Category ℓC ℓC') where
+  Slices : Categoryᴰ C (ℓ-max ℓC ℓC') ℓC'
+  Slices .Categoryᴰ.ob[_] c = Σ[ c' ∈ C .ob ] (C [ c' , c ])
+  Slices .Categoryᴰ.Hom[_][_,_] f P Q = Σ[ fᴰ ∈ C [ P .fst , Q .fst ] ]
+    f ∘⟨ C ⟩ P .snd  ≡ Q .snd ∘⟨ C ⟩ fᴰ
```
GenericMonkey, line 27: Interesting that the triangle becomes a square, but I can see how this gives us the pullback for free

Cubical/Categories/Displayed/Slice.agda
```diff
@@ -0,0 +1,52 @@
...
+      (λ _ → C .isSetHom _ _)
+      (C .⋆IdL (fᴰ .fst))
+  Slices .Categoryᴰ.⋆IdRᴰ fᴰ =
+    ΣPathPProp (λ _ → C .isSetHom _ _) (C .⋆IdR (fᴰ .fst))
+  Slices .Categoryᴰ.⋆Assocᴰ fᴰ gᴰ hᴰ =
+    ΣPathPProp (λ _ → C .isSetHom _ _) (C .⋆Assoc (fᴰ .fst) (gᴰ .fst) (hᴰ .fst))
+  Slices .Categoryᴰ.isSetHomᴰ =
+    isSetΣ (C .isSetHom) (λ f → isProp→isSet (C .isSetHom _ _))
+
+  Subobjects : Preorderᴰ C (ℓ-max ℓC ℓC') ℓC'
```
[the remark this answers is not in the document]
maxsnew: No it's not quotiented. It's a displayed poset if `C` is univalent. But is there any need to quotient if we work with displayed stuff? The only reason we wanted to quotient was so we could define it as a functor to POSET.

GenericMonkey (approved):
LGTM. Still need to spend some time thinking about what the displays in Quotient/Interp are actually capturing

Cubical/Categories/Displayed/Slice.agda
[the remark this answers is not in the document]
GenericMonkey: I guess maybe just naming then? My understanding was that Subobjects are defined as the iso classes of the monos.

> The only reason we wanted to quotient was so we could define it as a functor to POSET.

Do we not still want to do this in order to define the hyperdoctrine?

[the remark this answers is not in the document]
maxsnew: Well I don't think it's so clear cut as in the univalent case it's not necessary to quotient anything. For the hyperdoctrine stuff let's discuss on Zulip

Cubical/Categories/Displayed/Slice.agda
maxsnew: To be clear this doesn't give a pullback for free, it avoids the requirement of a pullback

PR #43, opened 2023-07-20
Adding workaround and refactor monotone

GenericMonkey:
- Adding the workaround from https://github.com/agda/cubical/issues/995#issuecomment-1630620766
- Refactoring definition of `MonFun` and eliminating the (now redundant) proofs/defns with explicit arguments

### DIFF
--- a/Cubical/Reflection/RecordEquiv/More.agda
+++ b/Cubical/Reflection/RecordEquiv/More.agda
@@ -1,0 +1,56 @@
+{-
+Adds the workaround for implicit arguments present in declareRecordIsoΣ
+from here: https://github.com/agda/cubical/issues/995
+Solution by @cmcmA20
+-}
+{-# OPTIONS --no-exact-split --safe #-}
+module Cubical.Reflection.RecordEquiv.More where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Isomorphism
+open import Cubical.Foundations.Equiv
+open import Cubical.Data.List as List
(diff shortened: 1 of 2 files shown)


Cubical/Reflection/RecordEquiv/More.agda
```diff
@@ -0,0 +1,56 @@
...
+
+  private variable Z : Bar
+
+  Baz : Bar → Type
+  Baz 0 = Unit
+  Baz _ = ℕ
+
+  record Foo : Type where
+    field
+      foo : Baz Z
```
maxsnew, line 53: A more thorough example would be better. Does the example work if `Foo` has parameters and if `foo` has parameters?

GenericMonkey:
```
module _ where private
  Bar : ℕ → Type
  Bar 0 = Unit
  Bar _ = ℕ

  record Foo (n : ℕ) : Type where
    field
      foo : {a : ℕ} → Bar a
      baz : Bar n

  Sigma : (n : ℕ) → Type
  Sigma n = Σ ({a : ℕ} → Bar a) (λ z → Bar n)

  module _ (x : ℕ) where
    foo-iso : Iso (Foo x) (Sigma x)
    unquoteDef foo-iso = defineRecordIsoΣ foo-iso (quote Foo)

```
Something like this?

maxsnew: Yes

PR #45, opened 2023-07-22
Displayed Universal Properties

maxsnew:
This is a start at defining universal properties for displayed categories in a similar manner to how we do it for ordinary categories. We define a displayed version of `SET`, displayed `Presheaf` over a presheaf, and a displayed universal element over a universal element. Then as specific examples, we define displayed right adjoints over a right adjoint and defined global and local binary and nullary products as well as cartesian lifts, which allows us to define when a displayed category is a fibration.

It's not quite lined up with what we are doing for categories because I don't have displayed profunctors or displayed functor comprehension yet, so I had to make a variant of RightAdjoint in the base to get things to line up definitionally for now.

### DIFF
--- a/Cubical/Categories/Adjoint/UniversalElements.agda
+++ b/Cubical/Categories/Adjoint/UniversalElements.agda
@@ -3,9 +3,13 @@
 module Cubical.Categories.Adjoint.UniversalElements where
 
 open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Isomorphism
 open import Cubical.Categories.Category
 open import Cubical.Categories.Functor
 open import Cubical.Categories.Profunctor.General
+open import Cubical.Categories.Presheaf.More
+open import Cubical.Categories.Presheaf.Representable
+open import Cubical.Categories.Instances.Sets
 
 private
   variable
@@ -23,6 +27,35 @@ RightAdjointAt : (C : Category ℓC ℓC')
                  (F : Functor C D)
                  (d : D .ob) → Type _
 RightAdjointAt C D F = UniversalElementAt D C (Functor→Profo-* C D F)
+
+-- Uh Oh
+RightAdjointAt' : (C : Category ℓC ℓC')
+                  (D : Category ℓD ℓD')
+                  (F : Functor C D) (d : D .ob)
+                → Type _
+RightAdjointAt' C D F d  =
+  UniversalElement C ((D [-, d ]) ∘F (F ^opF))
+
+RightAdjointAt→Prime : (C : Category ℓC ℓC')
+                 (D : Category ℓD ℓD')
+                 (F : Functor C D)
+                 (d : D .ob)
+                 → RightAdjointAt C D F d → RightAdjointAt' C D F d
+RightAdjointAt→Prime C D F d x .UniversalElement.vertex = UniversalElement.vertex x
+RightAdjointAt→Prime C D F d x .UniversalElement.element = UniversalElement.element x
+RightAdjointAt→Prime C D F d x .UniversalElement.universal = UniversalElement.universal x
+
+RightAdjoint' : (C : Category ℓC ℓC')
+                (D : Category ℓD ℓD')
+                (F : Functor C D)
+              → Type _
+RightAdjoint' C D F = ∀ d → RightAdjointAt' C D F d
+
+IdRightAdj' : (C : Category ℓC ℓC')
+      → RightAdjoint' C C Id
+IdRightAdj' C c .UniversalElement.vertex = c
+IdRightAdj' C c .UniversalElement.element = id C
+IdRightAdj' C c .UniversalElement.universal c' = isoToIsEquiv (iso _ (λ z → z) (C .⋆IdR) (C .⋆IdR))
 
 LeftAdjoint : (C : Category ℓC ℓC')
               (D : Category ℓD ℓD')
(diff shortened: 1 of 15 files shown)


Cubical/Categories/Adjoint/UniversalElements.agda
```diff
@@ -24,6 +28,39 @@ RightAdjointAt : (C : Category ℓC ℓC')
                  (d : D .ob) → Type _
 RightAdjointAt C D F = UniversalElementAt D C (Functor→Profo-* C D F)
 
+-- Uh Oh
```
stschaef, line 31: why Uh Oh?

Cubical/Categories/Displayed/Instances/Sets.agda
stschaef, line 1: It feels like it might be more natural to call this `SETᴰ` in line with the other defs

Cubical/Categories/Displayed/Presheaf.agda
stschaef, line 1: When defining a displayed universal element, what is exactly is the type of `elementᴰ`. A priori this seems to be a function out of `P .F-ob vertex` to some other fixed `hSet`, but which one?

Cubical/Categories/Displayed/Limits/Terminal.agda
```diff
@@ -0,0 +1,71 @@
...
+  module D = Categoryᴰ D
+  TerminalPresheafᴰ : (P : Presheaf C ℓP) → Presheafᴰ D P ℓ-zero
+  TerminalPresheafᴰ P .F-obᴰ x x₁ = Unit , isSetUnit
+  TerminalPresheafᴰ P .F-homᴰ = λ _ x _ → tt
+  TerminalPresheafᴰ P .F-idᴰ i = λ x x₁ → tt
+  TerminalPresheafᴰ P .F-seqᴰ fᴰ gᴰ i = λ x _ → tt
+
+  -- Terminal object over a terminal object
+  -- TODO: refactor using Constant Functorᴰ eventually
+  TerminalᴰSpec : Presheafᴰ D (TerminalPresheaf {C = C}) ℓ-zero
```
stschaef, line 44: What does `Spec` mean here

Cubical/Categories/Displayed/Limits/Terminal.agda
```diff
@@ -0,0 +1,71 @@
...
+open import Cubical.Categories.Displayed.Base
+open import Cubical.Categories.Displayed.Presheaf
+open import Cubical.Categories.Displayed.Functor
+open import Cubical.Categories.Limits.Terminal
+open import Cubical.Categories.Limits.Terminal.More
+open import Cubical.Categories.Presheaf.More
+
+-- There are multiple definitions of terminal object in a displayed category:
+-- 1. A terminal object in the total category, which is preserved by projection
+-- 2. A terminal object in the *fiber* of an object
```
stschaef, line 24: Being a lil shaky on how the limits below get instantiated, is 2. equivalent to this concrete description of a terminal object in a fiber? Reading Jacob's book asap

For `1 `your displayed terminal object over some `c`, for all `d : ob[c]`, `Hom[id][d, 1]` is contractible. If this is true, can you swap `id` out for any other endomorphism?

Cubical/Categories/Displayed/Limits/BinProduct.agda
stschaef, line 1: If you answer the terminal question that probably also answers this:

calling back to Thursday, we're still a little confused as to why these aren't the same type:

`ob[c ×C c']` where we take the product of objects in `C` first then go to the display; and, `ob[c] × ob[c']` where we take the nondependent product of the types spat out in the display. It seems like we get a map `ob[c] × ob[c']` to `ob[c ×C c']` via the universal property of the product, but why might this not cover all things in `ob[c ×C c']`

stschaef:
Pranav and I went through this together, so this is both of our reviews

Cubical/Categories/Displayed/Presheaf.agda
maxsnew: I don't understand the question, the type is `⟨ Pᴰ .F-obᴰ vertexᴰ element ⟩`

GenericMonkey: So our confusion was that `Pᴰ .F-obᴰ vertexᴰ` belongs to `SETS`, specifically to `ob [ P. F-ob vertex ]`. This type is defined as `⟨ P .F-ob vertex ⟩ → hSet ℓ'`. Thus, while it makes sense that this can now eat `element`, which is of type `⟨ P .F-ob vertex ⟩`, it's not clear what `hSet ℓ` the return type will be. Do we know anything about it?

GenericMonkey (approved):

stschaef (approved):

Cubical/Categories/Displayed/Limits/Terminal.agda
maxsnew: specification

Cubical/Categories/Displayed/Presheaf.agda
maxsnew: Resolved at the meeting

Cubical/Categories/Displayed/Limits/BinProduct.agda
maxsnew: > to why these aren't the same type

What does "these" refer to? Please try to be clearer in these questions.

Generally there are plenty of things in the fiber of `ob[ c xC c' ]` that are not a product of things from `ob[ c ]` and `ob[ c' ]`. For instance, if you take the displayed preorder of predicates on sets, then `c` and `c'` are sets and `ob[ c xC c' ]` is a subset of the product, i.e. a relation between `c` and `c'`. This has global products, given by taking a predicate `P` on `c` and `P'` on `c'` to the predicate `(P x P')(x, x') = P(x) /\ P'(x')`. But there are plenty of relations that don't arise in this way, for instance `Equal(x,x') = x == x'` is not of this form if `c` and `c'` are `Bool`.

PR #47, opened 2023-08-15
Alternative definition of Free Functor that avoids needing Eq

maxsnew:
@hejohns


maxsnew:
To get an `elim` version of this we need to implement https://github.com/maxsnew/multi-poly-cats/issues/46

hejohns (approved):

PR #48, opened 2023-08-29
Right adjoints in product category

stschaef:


maxsnew:
Does anyone remember what this was for? Specifically, the `UnivEltProd` file, the other is self explanatory

stschaef:
IIRC it was Eric's first pass at getting a universal element for the product presheaf. And then I came in later and try to generalize that construction as much as possible with the other file

I don't fully remember, but I think there was definitional differences between what he gave and what I made. Can't remember if they were reconcilable or not, can try to take a look at some point this week if you think its worth integrating both files

stschaef:
@ericgiovannini did you ever need this?

ericgiovannini:
> @ericgiovannini did you ever need this?

I haven't utilized it yet, but I expect that it will be useful for formulating limits in double categories

maxsnew:
What happened here?

PR #49, opened 2024-01-31
New free functor

maxsnew:
(Draft because I still need to clean some things up)

Give a new definition for the free functor between freely generated categories. Has the following differences from previous:
1. Based on quivers rather than graphs (better definitional behavior in some uses)
2. Freely builds in the functor rather than using a graph homomorphism
3. Defined using new Presented category construction
4. UMP is specified as an eliminator into a displayed category

Because of (2) this seems to avoid the need for any `Eq` manipulation in the definition of the eliminator. Need to port the solver to this to make sure the definitional behavior is right but on inspection it looks good.


maxsnew:
Ended up adding/changing a few more things, copying from the recent commit: 

1. Move Quiver to Data.Quiver, define Graph^D over Quiver, reformulate free category data in terms of Graph^D
2. Reformulate Presented Cat to take in a category rather than quiver
3. Use named module in presented cat

Cubical/Categories/Constructions/Free/Functor/AltPresented.agda
```diff
@@ -0,0 +1,138 @@
...
+          -- F⟪ id A ⟫ ≡ id (F ⟅ A ⟆)
+          satisfies-axioms (inl A) =
+            cong (𝓕 .F-homᴰ) (s .F-id)
+            ◁ 𝓕 .F-idᴰ
+          -- F⟪ f ⋆ g ⟫ ≡ F⟪ f ⟫ ⋆ F⟪ g ⟫
+          satisfies-axioms (inr (_ , _ , _ , f , g)) =
+            cong (𝓕 .F-homᴰ) (s .F-seq _ _)
+            ◁ 𝓕 .F-seqᴰ _ _
+
+module CoUnit {C : Category ℓc ℓc'} {D : Category ℓd ℓd'} (F : Functor C D)
```
stschaef, line 122: counit makes this sound like an adjunction, but I don't see what functors go into that adjunction. 

Is this just augmenting D by freely adding in the image of F, or is this also doing some adjunction business. Like also adding in the image of `FG d`, where `F : C -> D`, `G : D -> C` are adjoint.

stschaef: like a sort of free syntax to represent an adjoint to a given functor F. idk maybe I'm reading too much into this. I guess I just don't see if this is the free category you mention at the top of file, or if its an instance

Cubical/Categories/Displayed/Base/More.agda
```diff
@@ -61,3 +62,23 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'}
...
+  private
+    module C = Category C
+    module D = Category D
+
+  open Categoryᴰ Dᴰ
+  open Functor F
+  open Functorᴰ
+
+  -- TODO: better name?
+  reindexΠ : Functorᴰ F (reindex Dᴰ F) Dᴰ
```
stschaef, line 80: maybe "reindex" with a superscript "d" for displays? or maybe "reindexing-functor"?

Cubical/Categories/Displayed/Section.agda
```diff
@@ -102,6 +103,7 @@ module _ (C : Category ℓC ℓC') (P : Preorderᴰ C ℓD ℓD') where
     Section≡ _ (λ c → refl) (λ f → refl)
   PreorderSectionIsoCatSection .leftInv a = refl
 
+-- | TODO: this should go somewhere else
 module _ (C : Category ℓC ℓC') (D : Category ℓD ℓD') where
   weaken : Categoryᴰ C ℓD ℓD'
```
stschaef, line 108: why "weaken"

stschaef (commented):
Other than the couple naming questions above, this all looks good. 

Although, what cubical version are you on. On cubical HEAD, I seem to be getting issues loading these file stemming from something in `NaturalTransformation.Properties` or something

maxsnew:
> Although, what cubical version are you on. On cubical HEAD, I seem to be getting issues loading these file stemming from something in `NaturalTransformation.Properties` or something

You need to upgrade your version of Agda

Cubical/Categories/Constructions/Free/Functor/AltPresented.agda
maxsnew: This is the co-unit of a comonad on the 2-category of functors. I.e., for every functor F : C -> D we get a functor Free C - FreeFunctor > Free(C,D,F) and there is a square from this to the original functor. This gets used in the solver. I did use the name counit at least in a comment in the free category module (https://github.com/agda/cubical/blob/master/Cubical/Categories/Constructions/Free/Category/Base.agda#L159)

Cubical/Categories/Displayed/Section.agda
maxsnew: The idea is to think of a displayed category `C` over a category `G` like a dependent type `G |- C`. Then we can think of an ordinary category as dependent on an empty context `* |- C` and so this operation takes `* |- D` and turns it into `C |- wk(D)`, i.e., weakening.

Cubical/Categories/Displayed/Base/More.agda
maxsnew: Idk I think `reindexΠ` is more informative than both of those. In terms of the total space of the displayed categories involved, this is a projection.

stschaef: It seems opposite to some of the cubical naming scheme though. Like, `isSetΠ` and other similarly named things are refering to `Π` types

maxsnew: then maybe just lower-case pi?

maxsnew: I went with "forgetReindex" since it's a kind of forgetful functor

PR #59, opened 2024-03-02
Functor comprehension simplification

maxsnew:
This PR simplifies the FunctorComprehension proof, specifically the
part that constructs a representation from a universal element, in two
ways:

1. Use a principle `change-contractum` to get rid of some tedious
   reasoning about squares

2. Redefine the displayed category of representations so that its
   structure lines up more closely with the one for universal elements
   so there's fewer associativities

In the process I added a few more minor things:

1. several more useful constructions to displayed categories
2. cleaning up some stuff in the IsoComma HLevel proofs
3. define the FullImage of a functor (which I didn't end up needing).
4. proofs for when LiftF and postComposeF are FullyFaithful


stschaef (commented):
All looks good to me

PR #63, opened 2024-03-03
Prove SETᴰ is a fibration

maxsnew:
part of https://github.com/maxsnew/multi-poly-cats/issues/62


maxsnew:
Ok I pushed one more commit that does the reverse direction between hi tech and explicit notion of cartesian lifts. Should be good to go now @hejohns

PR #66, opened 2024-03-13
Create LICENSE

stschaef:
Add MIT License

Just listed "multi-poly-cats contributors" but we could put our names if you see fit


maxsnew:
LGTM. Merge it in after confirming that @GenericMonkey agrees to it

PR #67, opened 2024-03-17
Local sections

maxsnew:
Generalize the old definition of `Section`, which was a "global" section to be "local" sections. This seems much more ergonomic, and replaces many ugly uses of `Unit^D`. In general, `intro` forms for displayed categories should probably construct `Section`s rather than `Functor^D`s as it's easier to construct a `Functor^D` from a `Section` than vice-versa (see `Section→Functorᴰ` in `Categories.Constructions.TotalCategory`). I included some of @stschaef's upstream re-orgs as I was starting to get cyclic dependencies otherwise.


maxsnew:
TODO: make a version of elim for `Cubical.Categories.Constructions.Free.Category.Quiver` that uses `Reindex.Eq` rather than `Reindex`.

Cubical/Categories/Constructions/Quotient/More.agda
```diff
@@ -37,25 +37,26 @@ module _ (C : Category ℓ ℓ') where
...
-
-      -- TODO: make a macro that applies to Eq.refl in exactly this way
-      reindexᴰQuo : Categoryᴰ C ℓD ℓD'
-      reindexᴰQuo = reindex' Dᴰ QF Eq.refl λ _ _ → Eq.refl
+      module ReindexQuo = EqReindex Dᴰ QF Eq.refl (λ _ _ → Eq.refl)
 
       open Section
-      elim : (F : Section reindexᴰQuo)
+
+      -- TODO: should elim be the name for the global section or the local
```
stschaef, line 48: based on comments elsewhere, the default naming should be for local, right?

stschaef, line 48: based on comments elsewhere, the default naming should be for local, right?

Cubical/Categories/Displayed/Constructions/BinProduct/More.agda
```diff
@@ -38,8 +51,22 @@ module _ {C : Category ℓC ℓC'} (D : Categoryᴰ C ℓD ℓD')
 
 -- Product within a fiber
 -- TODO: replace this with ∫ᴰ ?
--- Product of total categories
--- todo: replace with the next construction & reindexing?
+
+-- TODO: is there an easy way to reduce this code duplication?
```
stschaef, line 55: whats's duplicated here? The module structure and parameters?

If so, can't these anonymous modules be nested to avoid duplication?

Cubical/Categories/Displayed/Constructions/Slice.agda
```diff
@@ -52,18 +56,11 @@ module _ (C : Category ℓC ℓC') (Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ') where
...
-      Fᴰ')
-    where
-      F = (∫F {Cᴰ = Cᴰ} (Wk.intro Id Id))
-      G = funcComp (Δ C) (Disp.Fst {Cᴰ = Cᴰ})
-      Gᴰ : Functorᴰ G (Unitᴰ (∫C Cᴰ)) (Hom C)
-      Gᴰ = (ID {C = C} ∘Fᴰ Unitᴰ.intro (Disp.Fst {Cᴰ = Cᴰ}))
-
-      Fᴰ' : Functorᴰ F (Unitᴰ (∫C Cᴰ)) (Hom C)
-      Fᴰ' = reindF' _ Eq.refl Eq.refl Gᴰ
+  Δ/C = TotalCatᴰ.introF _ _ (Wk.introF Id Id)
```
stschaef, line 59: I'm guessing that there is some gain in definitional behavior that motivates redefining this displayed slice. However, I don't necessarily see why

Cubical/Categories/Displayed/Section/Base.agda
```diff
@@ -0,0 +1,306 @@
...
+   Δ ----→ Γ
+       γ
+
+   Can be implemented as a Section (see
+   Displayed.Constructions.TotalCategory)
+
+            A
+          ↗ |
+         /  |
+      M /   |
```
stschaef, line 44: Killer diagrams in this file btw

Cubical/Data/Quiver/Base.agda
```diff
@@ -58,66 +59,59 @@ Quiver : ∀ ℓg ℓg' → Type _
...
-Cat→Quiver : Category ℓc ℓc' → Quiver ℓc (ℓ-max ℓc ℓc')
+Cat→Quiver : Category ℓC ℓC' → Quiver _ _
 Cat→Quiver 𝓒 = Graph→Quiver (Cat→Graph 𝓒)
 
-module _ (Q : Quiver ℓg ℓg') where
-  -- Of the four possible combinations: Q over Q, G over Q, Q over G,
-  -- Q over Q, this strikes the right balance. Having the displayed
-  -- thing be a Graph means we can define sections without using ≡.
-  record Graphᴰ ℓgᴰ ℓgᴰ' : Type (ℓ-max (ℓ-max ℓg ℓg') (ℓ-suc (ℓ-max ℓgᴰ ℓgᴰ')))
+-- A "heterogeneous" local section of a displayed graph
```
stschaef, line 99: heterogenous in what sense?

stschaef (commented):
Left some comments/question. I don't think I have any real critiques

There are a lot of changes, but up close this all looks pretty sensible.

The only high level question I have is about how this might relate to the existing cubical repo --- i.e. do you foresee this being merged in? In working with these displayed categories, we've seen that there is good reason to use `Data.Equality` over a path type. However, it also seems like the cubical repo has a preference for using the the cubical path type by default over the identity type when possible. In that, refactors like this maybe live outside of their "programming program". Maybe this is concern is misplaced though

Cubical/Categories/Displayed/Section/Base.agda
hejohns: Sanity checking myself: the variables in the diagrams and the variables in the definitions aren't the same, right?

hejohns: (The diagram variable names are presumably contexts/dependent types/context extensions?)

hejohns: (While the definitions are in terms of standard non-logic variable names.)
This unfortunately turned into three comments.

Cubical/Categories/Displayed/Section/Base.agda
```diff
@@ -0,0 +1,306 @@
...
+
+   Style is offer both: elim to construct a local section and elim' to
+   construct a global section.
+-}
+
+module _ {C : Category ℓC ℓC'}
+         (Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ')
+         where
+  GlobalSection : Type _
+  GlobalSection = Section Id Cᴰ
```
hejohns, line 152: I may find the answer as I keep reading, but I still don't get the global v. local terminology here.

hejohns: A global section corresponds to $$\bar{x} : Γ ⊢ M(\bar{x}) : A(\bar{x})$$
while a local section corresponds to $$\bar{z} : Δ ⊢ M(\bar{z}) : A(γ(\bar{z}))$$ or equivalently $$\bar{z} : Δ ⊢ M(\bar{z}) : γ^*(A(\bar{x}))$$? In what sense is one global and the other local?

hejohns: I suppose it's just that the local section is only a section over "a subset" (very loosely construed) of A.
(Thinking of [fiber bundles](https://en.wikipedia.org/wiki/Section_(fiber_bundle))).

I guess that's it.

Cubical/Categories/Displayed/Constructions/BinProduct/More.agda
maxsnew: Oh this comment is outdated because I fixed the code duplication

Cubical/Data/Quiver/Base.agda
maxsnew: bc it's a graph over a quiver rather than a quiver over a quiver or graph over a graph

Cubical/Categories/Constructions/Quotient/More.agda
maxsnew: I thought so but the local `elim`s get you into transport hell...

Cubical/Categories/Displayed/Section/Base.agda
maxsnew: you can edit comments lol

maxsnew: And yes I used context and context extension syntax here. I guess I shouldn't

maxsnew:
> The only high level question I have is about how this might relate to the existing cubical repo --- i.e. do you foresee this being merged in?

I really don't know but they are so slow processing PRs that it is nearly indistinguishable with none of this being merged in.

Cubical/Categories/Displayed/Constructions/Slice.agda
maxsnew: I don't really understand the question. This construction isn't really being changed at all in this PR, it's just being updated because its dependencies changed.

As for where it's used, it's in the definition of a Fibration.

Cubical/Categories/Constructions/Quotient/More.agda
maxsnew: I think what I'll do is that `elim` will mean a global section (which matches how `elim` for e.g., sums works) and call the other one `elimLocal`

Issue #89, opened 2024-05-19
incremental Makefile


maxsnew:
current Makefile is not incremental at all, it recompiles everything every time. I tried it out and the one from `agda-categories` does not have this issue so we should copy it: https://github.com/agda/agda-categories/blob/master/Makefile

hejohns:
Reading the zulip message made me think you were going to propose changing the Makefile to look for agdai files under ./_build/.
Double checking that you do in fact mean scripting (well, copying the script from agda-cateogories) a top level "import everything" agda file, and just invoking Make to check that one file, letting agda handle the caching.

Also, the current Makefile only checks ./Cubical/.
Is this something we still want to do?

Otherwise I can do this right now.

maxsnew:
I think we should just adapt the approach taken in their script. And it should build everything not just Cubical

maxsnew:
implemented by https://github.com/maxsnew/cubical-categorical-logic/pull/90

PR #90, opened 2024-05-19
incremental Makefile

hejohns:
#89


hejohns:
@maxsnew, what do you want to do about UMP.agda? It doesn't build right now.

maxsnew:
Delete it but make an issue for us to reimplement what it had (which I think is really just preservation of products)

maxsnew:
the sed script does not work on Mac OS X

hejohns:
Oh I didn’t expect the macos basic regex to be slightly different than gnu
sed.
Will fix.

maxsnew:
confirmed working on my machine:
```
> make clean
...
> time make
...
make  103.86s user 2.93s system 97% cpu 1:49.09 total
> time make
...
make  3.40s user 0.34s system 93% cpu 3.993 total
```

Thanks @hejohns !

PR #92, opened 2024-05-20
Relative monad

maxsnew:
Some theory of relative monads, as well as the notion of a displayed category lifting limits. Includes the formulation of the forgetful functor from algebras lifting limits but not the construction yet.


maxsnew:
Closing as low-priority right now.

Issue #85, opened 2024-05-17
Move CI line length check to a tracked script?


hejohns:
@stschaef 

Is there a reason the CI script to check line lengths is inlined directly in the yaml file?
https://github.com/maxsnew/cubical-categorical-logic/blob/41323e59ca745ce33f76f1e6ff2c7c8357b701b2/.github/workflows/main.yml#L86

I've copy/pasted it for reference:

```
#!/usr/bin/env bash

error=0
shopt -s globstar nullglob
for file in **/*.agda; do
  python3 -c '
import sys
from wcwidth import wcswidth
filename = sys.argv[1]
try:
    with open(filename, "r", encoding="utf-8") as file:
        for lineno, line in enumerate(file, 1):
            if wcswidth(line.rstrip("\n")) > 80:
                print(f"{filename}:{lineno}: line too long")
                sys.exit(1)
except Exception as e:
    print(f"Error processing {filename}: {e}")
    sys.exit(1)
  ' "${file}" || error=1
done
exit ${error}
```
What did you mean by you haven't written anything for checking locally yet?
https://github.com/maxsnew/cubical-categorical-logic/blob/41323e59ca745ce33f76f1e6ff2c7c8357b701b2/README.org?plain=1#L13
This works perfectly fine locally?

Can we move it into sh file, add a Makefile target for checking line lengths, and have the CI run eg `make check-line-lengths` to call the script?

I vaguely remember you mentioned something about this a year ago.

stschaef:
My first pass through this was just a quick and dirty rip off of some of cubical's CI. So the answer to your questions is "no, there is no good reason" 


What you've added looks great! It seems like #93 concludes this issue. If not, are there any other CI related things I should take a look at

hejohns:
There was also https://github.com/maxsnew/cubical-categorical-logic/issues/84

PR #93, opened 2024-05-20
move line length check to script

hejohns:
https://github.com/maxsnew/cubical-categorical-logic/issues/85


stschaef (approved):
Looks good to me!

PR #95, opened 2024-05-22
Gluing for Free cat with terminal objects

maxsnew:
I got started on implementing some metatheory for free categories with terminal objects. I got up to constructing the eliminator that constructs a local section but see we actually don't have the proof that VerticalTerminal is preserved by reindexing or that it implies LiftedTerminal. @hejohns please add those to this PR


hejohns:
Not these?

https://github.com/maxsnew/cubical-categorical-logic/blob/6bb2d8a2af861f5757eb287d598f413d921a2788/Cubical/Categories/Displayed/Fibration/Properties.agda#L86

https://github.com/maxsnew/cubical-categorical-logic/blob/6bb2d8a2af861f5757eb287d598f413d921a2788/Cubical/Categories/Displayed/Fibration/Properties.agda#L109

maxsnew:
Well they don't have anything to do with fibrations so why would I look there

hejohns:
I can move these, but I had them under Fibration/ for uniformity.
Or at least Vertical→Lifted in general should be under Fibration/.
I agree (not that you're necessarily saying this) that preservation under reindexing should get moved to Displayed/Limit/Terminal

maxsnew:
No all of these constructions should be under
Constructions.Reindex.Properties

maxsnew:
If they're already there I'll move them myself

hejohns (approved):
Looks good to merge.
I think I follow how we want to do these gluing examples.

I don't get at the moment why the uniqueness of normal forms has a path issue while the uniqueness of canonical forms is just fine, but maybe I'll get some idea after doing canonical forms for free cartesian categories.

maxsnew:
> I don't get at the moment why the uniqueness of normal forms has a path issue while the uniqueness of canonical forms is just fine, but maybe I'll get some idea after doing canonical forms for free cartesian categories.

I think it's just `cubical` nonsense because the NFs are a more complicated data type than sums. It could be a bug in cubical Agda for all we know

PR #96, opened 2024-05-23
SETᴰ has VerticalTerminalsᴰ

hejohns:
As per zulip messages about https://github.com/maxsnew/cubical-categorical-logic/pull/95

Using the current module locations of Vertical->Lifted and such, which are moving in the above PR.


maxsnew (approved):
lgtm

Issue #54, opened 2024-02-19
Wide Subcategories


maxsnew:
We defined [full subcategories](https://github.com/maxsnew/multi-poly-cats/blob/new-profunctors/Cubical/Categories/Displayed/Constructions/FullSubcategory.agda) as displayed categories in the new PR, it would be nice to have the opposite extreme, a ["wide" subcategory](https://ncatlab.org/nlab/show/wide+subcategory) as well.

maxsnew:
Naming update: `FullSubcategory` has been merged upstream and is now called `PropertyOver`. I guess that means `WideSubcategory` should be like `HomPropertyOver` or something

stschaef:
See #100

PR #100, opened 2024-06-03
Wide subcategories

nvarner:
Defines wide subcategories (`HomPropertyOver`) for #54. Includes some basic examples, and incorporates some preliminary feedback from @stschaef.


Cubical/Categories/Displayed/Constructions/HomPropertyOver.agda
```diff
@@ -0,0 +1,113 @@
...
+    ℓC ℓC' ℓP : Level
+
+open Category
+
+module _
+  (C : Category ℓC ℓC') (P : ∀ {x y} → C [ x , y ] → Type ℓP)
+  (Pprop : ∀ {x y} (f : C [ x , y ]) → isProp (P f))
+  (Pid : ∀ {x} → P (id C {x}))
+  (Pcomp : ∀ {x y z} (f : C [ x , y ]) (g : C [ y , z ]) → P f → P g → P (_⋆_ C f g))
+  where
```
maxsnew, line 22: I think there's so many fields here that we should make this a struct `HomPropertyOver` like we do with `StructureOver`.

maxsnew (changes requested):
Looks great but I think we should make it a struct like we do with StructureOver

Cubical/Categories/Displayed/Constructions/HomPropertyOver.agda
nvarner: I've changed `HomPropertyOver` to a record and created a defined to displayed categories.

nvarner: I've updated these names to match, except for `Hom[_]`. It caused a parser ambiguity between `Hom[_,_]` and `Hom[ (_,_) ]`. I used `Hom[_][-,-]` instead, which is the most reasonable idea I've had.

PR #102, opened 2024-07-11
Presheaf ccc

bond15:


bond15:
Still cleaning things up, but there are no holes left.

bond15:
Trying to add all the new structure to Cubical.Categories.Preseheaf.Constructions .. 
but there is an annoying cyclic module dependency
  Cubical.Categories.Presheaf.Constructions
  Cubical.Categories.Exponentials
  Cubical.Categories.Limits.BinProduct.More
  Cubical.Categories.Presheaf.Constructions

hejohns (commented):
nice! It looks like line lengths/whitespace are the only things failing the CI.

maxsnew:
To fix the whitespace you can run the `fix-whitespace` utility. We should have instructions on how to install it in the readme but we don't seem to...
For the long line lengths, I added `(add-hook 'agda2-mode-hook 'column-enforce-mode)` to my `init.el`.

maxsnew:
I haven't looked at the code yet, but `Presheaf.CCC` is not the typical naming scheme for this. It would probably go in `Presheaf.Properties` if that doesn't cause a circular dependency or just add it to the existing `Presheaf.Constructions`

hejohns:
I tweaked the universe levels for $\top\mathcal P$ and $\times\mathcal P$ to simply be over $\mathsf{SET} \ell S$-valued presheaves, and moved them into a separate module from $\Rightarrow\mathcal P$ which *does* need the level of $\mathsf{SET}$ to be "high enough".

hejohns:
note after meeting w/ @maxsnew:
should make presheaf operations more universe polymorphic, eg should be able to take a product of presheaves at $\ell A$ and $\ell B$ for a presheaf at $\ell \mathsf{-max}\ \ell A\ \ell B$

bond15:
I missed these notifications. Looking into levels issue

bond15:
I made the `PshProd` and `ExpOb` more universe polymorphic but I've run into an issue.

https://github.com/maxsnew/cubical-categorical-logic/blob/412ac9858ee4e57a5318b0ff695394fc0bd26e57/Cubical/Categories/Presheaf/CCC.agda#L133

The last remaining hole here is `refl` .. you can normalize the goal and lhs and rhs are exactly the same.
If you fill the hole with refl and attempt to typecheck, Agda runs out of memory.

maxsnew:
Why bother doing a comp with `refl`? Why not just
```
h .N-ob y (Z .F-hom (y→x .lower) Zx , By)) (sym prf) ) 
```

maxsnew:
Why bother doing a comp with `refl`? Why not just
```
 liftExt (cong (λ h → h .N-ob y (Z .F-hom (y→x .lower) Zx , By)) (sym prf)) 
```

maxsnew:
I rebased onto main so make sure you reset locally

PR #101, opened 2024-07-02
Conservativity of freely adding 1,× to a free Category

hejohns:
WTS that over a Quiver Q, the inclusion of the free category over Q into the free cartesian category over Q (as a trivial ProductQuiver) is fully faithful.

Faithfulness is quick, but we're going to use a gluing normalization argument (over the free cartesian category) for fullness.

WIP: a very concrete displayed category of presheaves, given directly, eg not as slices, not as a displayed category of displayed functors and natural transformations

PR #106, opened 2024-08-12
Add Makefile support for .lagda.md and make Gluing/Conservativity.agda (#101) literate

hejohns:


hejohns:
The html support was already added by @stschaef in #144, and I no longer see the need to add lagda support at the moment.

PR #115, opened 2024-10-25
Monoidal combinators

maxsnew:


stschaef (approved):
All looks good to me

I'm surprised left whiskering of a `NatIso` made it upstream but not right whiskering

PR #119, opened 2025-02-26
Put caching back into the CI

stschaef:
Refactored the CI workflow to more closely mirror `agda/cubical`

Builds now don't do any redundant work, including in code from `cubical`. This is effectively the same strategy as in `grammars-and-semantic-actions`

https://github.com/maxsnew/grammars-and-semantic-actions/pull/32


maxsnew:
lgtm

PR #123, opened 2025-07-14
Displayed functor comprehension

maxsnew:
I figured out that we can reuse a lot of theorems on ordinary universal properties for displayed universal properties using total categories. The key ingredients here are the definitions of ΣF, ∫F-Functor, ∫Prof, ∫UE and ∫UEs. This means we can for instance derive the displayed universal element notation essentially unchanged from the ordinary universal elements, plus the displayed functor comprehension can be easily extracted from the ordinary functor comphrehension.

I also included some nice notation for displayed presheaves that we should probably back-port to ordinary presheaves.


hejohns (approved):

hejohns:
@stschaef (I understand what you were saying about functor comprehension now, much later)

Issue #122, opened 2025-07-14
Displayed Exponentials


maxsnew:

maxsnew:
Implemented in https://github.com/maxsnew/cubical-categorical-logic/pull/125

PR #125, opened 2025-07-15
Displayed exponentials, Simpler Functor Comprehension, Displayed Bifunctors, PresheafNotation

maxsnew:
Subsumes (https://github.com/maxsnew/cubical-categorical-logic/pull/124).

This PR is to defined displayed Exponentials (closing https://github.com/maxsnew/cubical-categorical-logic/issues/122). In the process it defines the minimal amount of displayed infrastructure to display over the definition of exponentials. I ended up tweaking some definitions of binary products with a fixed object to minimize the amount I needed to define without losing any of the nice definitional behavior we worked so hard on.

There's also some QoL stuff. Mainly a direct proof of FunctorComprehension that is much faster to type check and some extensions to PresheafNotation and Presheaf^DNotation, though I haven't really needed Presheaf^DNotation as much as I thought I would because we get most things directly from \int P. I also made a bunch of the proofs in PresheafNotation and Presheaf^DNotation opaque since I assume that can only make type checking faster.

### DIFF
--- a/Cubical/Categories/Displayed/Presheaf.agda
+++ b/Cubical/Categories/Displayed/Presheaf.agda
@@ -36,6 +36,8 @@ open Functorᴰ
 
 -- equivalent to the data of a presheaf Pᴰ over ∫ D and a natural transformation
 -- Pᴰ → P ∘ Fst
+--
+-- IMO the order of D and P here should be swapped to match Functorᴰ
 Presheafᴰ : {C : Category ℓC ℓC'} (D : Categoryᴰ C ℓD ℓD')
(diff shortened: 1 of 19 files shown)


Cubical/Categories/Displayed/Presheaf.agda
```diff
@@ -36,6 +36,8 @@ open Functorᴰ
 
 -- equivalent to the data of a presheaf Pᴰ over ∫ D and a natural transformation
 -- Pᴰ → P ∘ Fst
+--
+-- IMO the order of D and P here should be swapped to match Functorᴰ
```
hejohns, line 40: Sure, could be a small refactor issue

Issue #128, opened 2025-07-15
Universal Quantifier/Pi


maxsnew:
Should be able to define this as some kind of right adjoint to taking the cartesian lift. Will probably need some kind of functor comprehension nonsense to define "taking a cartesian lift" as a functor.

maxsnew:
We don't really need a full bifibration, we just need the universal quantifier/pi for displayed objects. I'm suspicious that it's impossible to define a full bifibration the way we are doing things in the usual way of a right adjoint, so that's probably a dead end.

The universal property for Π should be `C [ f ][ wᴰ , Π zᴰ ] ≅ C [ π₂ ⋆ᴰ f ][ π₁* wᴰ , zᴰ ]` where here `zᴰ : Cᴰ.ob[ x × y ]` and `Π zᴰ : Cᴰ.ob[ y ]`.

maxsnew:
Ok here's how to define it as a right adjoint. For a fixed object y : C, if all products with y exist you get a functor -×y : C → C. Then given a displayed category Cᴰ you can reindex it -×y* Cᴰ (still displayed over C). An object of -×y* Cᴰ over x is Cᴰ.ob[ x × y ].

Then you can define a vertical functor from Cᴰ to -×y* Cᴰ mapping an object xᴰ over x to π₁* x. Universal quantification/Π is then right adjoint to that.

So the difficulty in *defining* Π is just in defining that vertical functor π₁* : Cᴰ → -×y* Cᴰ . If you use the right-adjoint formulation of cartesian lifts then we have all of the displayed functor combinators you need to define this. It will probably be very slow to type check but it should work.

maxsnew:
You also get the definition of existential quantifier/weak sigma out of this, as a left adjoint.

PR #132, opened 2025-07-21
Universal quantifier, functorial cartesian lifts, vertical universal elements stuff, some re-orgs

maxsnew:
Cosmetic changes:
1. Removed old Bifunctor.agda and replaced it with what used to be called Bifunctor.Redundant (long overdue change)
2. Rename VerticalRightAdjointᴰ to RightAdjointⱽ

More substantive changes:
1. Implement some Presheafⱽ notation and UniversalElementⱽ stuff
3. Define cartesian lifts as universal elements of a manually defined profunctor out of the displayed slice category. The simpler definition using right adjoints seems too slow in practice.
4. Define universal quantifiers as right adjoint to taking the cartesian lift of projection. This implements https://github.com/maxsnew/cubical-categorical-logic/issues/128


maxsnew:
Currently I did leave a bit of a mismatch between the functorial definition of cartesian lift and the old one we had defined manually. Current gluing proofs use the manual one, and the manual one is slightly nicer, so we should keep it around IMO and prove they are equivalent.

PR #147, opened 2025-08-06
Vertical Exponentials for Displayed Sets

stschaef:
To do the equational proofs, it took me two days of unpacking equivalent definitions to realize how simple this actually is

The code can be cleaned up and compacted, but I'm gonna submit this PR and go home bc I've been staring at my monitor for far too long

It seems that a lot of things are definitionally nice in displayed sets. It's been a while since using them that I'd forgotten that `f` composed with identity is definitionally `f`, same with associativity. I was also surprised that the cartesian lift definitionally preserves the vertex of an exponential in each fiber category
```
    vert≡ : ∀ {Aᴰ Aᴰ'} → (f* (Aᴰ ⇒A Aᴰ')) ≡ ((f* Aᴰ) ⇒B (f* Aᴰ'))
    vert≡ = refl
```

In an arbitrary category, I'd expect exponential objects should only be preserved up to iso, is this correct? (I'm trying to understand how much harder the presheaf example might be than this)

### DIFF
--- a/Cubical/Categories/Displayed/Instances/Sets/Properties.agda
+++ b/Cubical/Categories/Displayed/Instances/Sets/Properties.agda
@@ -119,43 +119,34 @@ module _ {ℓ} {ℓ'} where
   FiberExponentialSETᴰ A Aᴰ Aᴰ' .vertex a .fst = ⟨ Aᴰ a ⟩ → ⟨ Aᴰ' a ⟩
   FiberExponentialSETᴰ A Aᴰ Aᴰ' .vertex a .snd = isSet→ (str (Aᴰ' a))
   FiberExponentialSETᴰ A Aᴰ Aᴰ' .element a (f , aᴰ) = f aᴰ
-  FiberExponentialSETᴰ A Aᴰ Aᴰ' .universal Aᴰ'' .equiv-proof f .fst .fst a aᴰ'' aᴰ =
-    f a (aᴰ'' , aᴰ)
-  FiberExponentialSETᴰ A Aᴰ Aᴰ' .universal Aᴰ'' .equiv-proof f .fst .snd =
-    fromPathP (λ i →
-      transport-filler (λ j → (a : ⟨ A ⟩) → ⟨ Aᴰ'' a ⟩ × ⟨ Aᴰ a ⟩ → ⟨ Aᴰ' a ⟩)
-                        f (~ i)
-    )
-  FiberExponentialSETᴰ A Aᴰ Aᴰ' .universal Aᴰ'' .equiv-proof f .snd (g , x) =
-    ΣPathP (
-      funExt₂ (λ a aᴰ'' → funExt λ aᴰ → sym (funExt⁻ (funExt⁻ x a) (aᴰ'' , aᴰ)))
-      ∙ fromPathP (λ i →
-          transport-filler (λ j → (a : ⟨ A ⟩) → ⟨ Aᴰ'' a ⟩ → ⟨ Aᴰ a ⟩ → ⟨ Aᴰ' a ⟩)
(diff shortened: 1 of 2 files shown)


stschaef:
It's cleaned up now and ready for review

Cubical/Categories/Displayed/Instances/Sets/Properties.agda
```diff
@@ -85,3 +101,67 @@ SETᴰCartesianCategoryⱽ ℓ ℓ' .CartesianCategoryⱽ.Cᴰ = SETᴰ ℓ ℓ'
...
+    bpw {A = A} Aᴰ Aᴰ' = bp A (Aᴰ' , Aᴰ)
+
+  open Functor
+  open UniversalElement
+  FiberExponentialSETᴰ : (A : SET ℓ .ob) → (Aᴰ Aᴰ' : SETᴰ.ob[ A ]) →
+    Exponential SETᴰ.v[ A ] Aᴰ Aᴰ' (bpw Aᴰ)
+  FiberExponentialSETᴰ A Aᴰ Aᴰ' .vertex a .fst = ⟨ Aᴰ a ⟩ → ⟨ Aᴰ' a ⟩
+  FiberExponentialSETᴰ A Aᴰ Aᴰ' .vertex a .snd = isSet→ (str (Aᴰ' a))
+  FiberExponentialSETᴰ A Aᴰ Aᴰ' .element a (f , aᴰ) = f aᴰ
+  FiberExponentialSETᴰ A Aᴰ Aᴰ' .universal Aᴰ'' .equiv-proof f .fst .fst a aᴰ'' aᴰ =
```
maxsnew, line 122: I find using `isIsoToIsEquiv` to be easier

maxsnew, line 122: I find using `isIsoToIsEquiv` to be more natural.

Cubical/Categories/Exponentials.agda
```diff
@@ -94,6 +94,9 @@ module ExponentialNotation {C : Category ℓC ℓC'}{c d} -×c (exp : Exponentia
   lda : ∀ {Γ} → C [ Γ ×a , d ] → C [ Γ , vert ]
   lda = intro
 
+  β⇒ : ∀ {Γ} → (f : C [ Γ ×a , d ]) → ((π₁ C.⋆ lda f) ,p π₂) C.⋆ app ≡ f
```
maxsnew, line 97: Not sure if we need this, isn't `β` already exported?

stschaef: It is already exported, but I like reexporting specific instances of `β` for each of exponentials, binproducts, etc, because it makes it clear what the signature of `β` is for each construction

It took me a frustratingly long time to recognize that `β` was definitionally the above equality, so I think it's useful to keep redundant but readable definitions like this

maxsnew: The one for binproducts is different though: xβ1 and xβ2 are not β themselves, they are applications of cong fst and cong snd to β. I think discoverability is an issue, but I think the best name for this is β. If you want to explicitly document it in the code, you can import the original β qualified and then define β to be equal to it.

stschaef: Oh wait, I may have misunderstood your comment

I took it to mean that I'd import `β` qualified in the usage site at `Instances.Sets.Properties`, in which case its easy to just remove it

But i think you meant that I could re-export the name from `Exponentials`. Ah, anyway, I think what's pushed now should be fine

maxsnew:
I simplified it a bit further, to show what I meant. Also got rid of a lot of annotations that were inferrable.

PR #157, opened 2025-08-14
Category with Families and the Simple variant

maxsnew:
I also included Σ for presheaves, but didn't need it yet.

### DIFF
--- a/Cubical/Categories/Displayed/Constructions/Reindex.agda
+++ b/Cubical/Categories/Displayed/Constructions/Reindex.agda
@@ -1,0 +1,6 @@
+{-# OPTIONS --safe #-}
+module Cubical.Categories.Displayed.Constructions.Reindex where
+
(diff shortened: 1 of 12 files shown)


maxsnew:
This PR has a definition of Simple Categories with Families as well as vertical and displayed versions and the full theory for gluing (with only base types). There's some really interesting new concepts here that I haven't seen before but were very natural for generalizing from Categories to SCwF. The main one is `Displayed.Presheaf.CartesianLift` which is a generalization of a cartesian lift to presheaves. This is a perfect generalization in that there is a definitional isomorphism between the definition of `CartesianLift` we already had and a Presheaf CartesianLift for a representable. This suggests to me that we should port everything to use the presheaf version. For the vertical to displayed construction I also had to generalize the proof that vertical products + fibration implies displayed products to the simple comprehension. Again this was a perfect generalization and I was able to just port the old proof to work with presheaves that are not assumed to be representable and instantiate it to get the proof for cartesian categories.

PR #165, opened 2025-08-22
Universe Polymorphic Presheaf Hom,Iso and Displayed Variants. Path-based + Iso based vertical to displayed products/terminals

maxsnew:
This is a pretty big PR that gives us two new methods for carrying out our constructions on displayed universal elements. The basic idea is that most of the work should not involve universal elements specifically, but reason directly about the (displayed) presheaves involved. With univalence, you can then work directly with paths, and the nice thing about that is you can just use the existing combinators for Path/PathP composition. Unfortunately, this is not universe polymorphic enough for the SCwF proofs. Rather than futz with lifts, I implemented a universe-polymorphic theory of (displayed) presheaf homomorphisms and isomorphisms, and their composition with each other and universal elements. Additionally there are some constructions of presheaves and homomorphisms, especially to do with products, terminals and reindexing.

I'm mostly looking for feedback on the design/naming of the new utilities. I think I ended up not needing everything so I might just take out some of the messier stuff like the `Representation` types that maybe aren't really useful on their own.

### DIFF
--- a/Cubical/Categories/Displayed/Presheaf/Constructions.agda
+++ b/Cubical/Categories/Displayed/Presheaf/Constructions.agda
@@ -174,6 +174,27 @@ module _ {C : Category ℓ ℓ'} {Cᴰ : Categoryᴰ C ℓᴰ ℓᴰ'}
   PshProdⱽ≅ᴰ .snd .rightInv _ _ = refl
   PshProdⱽ≅ᴰ .snd .leftInv _ _ = refl
 
+module _ {C : Category ℓC ℓC'}{Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'}
+  {P : Presheaf C ℓP}{Q : Presheaf C ℓQ}
+  {α β : PshHom P Q}(α≡β : α ≡ β) {Qᴰ : Presheafᴰ Q Cᴰ ℓQᴰ}
+  where
+  private
+    module Qᴰ = PresheafᴰNotation Qᴰ
+  reindPathIsoⱽ : PshIsoⱽ (reind α Qᴰ) (reind β Qᴰ)
+  reindPathIsoⱽ .fst .PshHomᴰ.N-obᴰ = Qᴰ.reind (funExt⁻ (funExt⁻ (cong fst α≡β) _) _)
+  reindPathIsoⱽ .fst .PshHomᴰ.N-homᴰ =
+    Qᴰ.rectify $ Qᴰ.≡out $
+      (sym (Qᴰ.reind-filler _ _)
+      ∙ sym (Qᴰ.reind-filler _ _)
+      ∙ Qᴰ.⟨⟩⋆⟨ Qᴰ.reind-filler _ _ ⟩)
+      ∙ Qᴰ.reind-filler _ _
+  reindPathIsoⱽ .snd .isIsoOver.inv q =
+    Qᴰ.reind ((funExt⁻ (funExt⁻ (cong fst (sym α≡β)) _) _))
+  reindPathIsoⱽ .snd .isIsoOver.rightInv q qᴰ =
+    Qᴰ.rectify $ Qᴰ.≡out $ sym $ Qᴰ.reind-filler _ _ ∙ Qᴰ.reind-filler _ _
+  reindPathIsoⱽ .snd .isIsoOver.leftInv q qᴰ =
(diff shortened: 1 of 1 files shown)


Cubical/Categories/Displayed/Presheaf/CartesianLift.agda
```diff
@@ -152,6 +190,12 @@ module _ {C : Category ℓC ℓC'} {Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'}
   YoFibrations→isCatFibration : YoFibrations → Fibration.isFibration Cᴰ
   YoFibrations→isCatFibration YoLifts cᴰ' f = YoLift→CatLift $ YoLifts cᴰ' f
 
+module _ {C : Category ℓC ℓC'} (Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ') where
+  private
+    module Cᴰ = Categoryᴰ Cᴰ
+  isCatFibration : Type _
+  isCatFibration = ∀ {x} (xᴰ : Cᴰ.ob[ x ]) → isFibration' (Cᴰ [-][-, xᴰ ])
```
stschaef, line 197: I don't understand the two separate usages of the term `CatFibration` here. You use it at `L190` for `Fibration.isFibration` and at `L196` for `isFibration'` of representables

I would also say that the defintion at `L196` should probably be `isCatFibration'` for consistency, because it uses `isFibration'`

Cubical/Categories/Instances/Sets/More.agda
```diff
@@ -39,3 +43,66 @@ open Functor
...
+  CatIso→Iso' : Iso ⟨ A ⟩ ⟨ B ⟩
+  CatIso→Iso' = isHAEquiv→Iso (CatIso→HAE .snd)
+
+  CatIso→⟨A⟩≡⟨B⟩ : ⟨ A ⟩ ≡ ⟨ B ⟩
+  CatIso→⟨A⟩≡⟨B⟩ = ua (_ , (isHAEquiv→isEquiv (CatIso→HAE .snd)))
+
+  CatIso→A≡B : A ≡ B
+  CatIso→A≡B = Σ≡Prop (λ _ → isPropIsSet) CatIso→⟨A⟩≡⟨B⟩
+
+isUnivalentSET' : isUnivalent (SET ℓ)
```
stschaef, line 80: Why do you need a different proof of univalence for sets than the one thats in the library currently?

stschaef: I'm assuming its because of the definitional test below? That is, you want `ua` to output the correct thing?

stschaef (commented):
This is a huge PR, so I haven't commented much on the `PshHom` stuff because that looked pretty straightforward. I left a few comments and questions, but will probably need a closer look again after lunch

Cubical/Categories/Displayed/Limits/Cartesian.agda
```diff
@@ -50,3 +68,42 @@ module _ {CC : CartesianCategory ℓC ℓC'}
...
+
+record CartesianCategoryReprⱽ (C : Category ℓC ℓC') (ℓCᴰ ℓCᴰ' : Level)
+  : Type (ℓ-suc (ℓ-max ℓC (ℓ-max ℓC' (ℓ-max ℓCᴰ ℓCᴰ')))) where
+  no-eta-equality
+  field
+    Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'
+  module Cᴰ = Categoryᴰ Cᴰ
+  field
+    termⱽ : ∀ {c} → Representationᵁⱽ Cᴰ (TerminalPresheafᴰ* Cᴰ ℓCᴰ' (C [-, c ]))
+    bpⱽ   : ∀ {c} (cᴰ dᴰ : Cᴰ.ob[ c ]) → Representationᵁⱽ Cᴰ ((Cᴰ [-][-, cᴰ ]) ×ⱽPsh (Cᴰ [-][-, dᴰ ]))
```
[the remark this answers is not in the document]
maxsnew: Nah, I think `C [-, c ]` is already the most common thing in the library so we should use it and things that look like it.

Cubical/Categories/Instances/Sets/More.agda
maxsnew: yes, it's nice if applying `isUnivalentSet` in the way I described actually behaves the same as applying the univalence axiom directly. The version upstream has unnecessary transports. I added a comment so we remember to upstream it later

PR #171, opened 2025-08-26
Eq-based PshIsoⱽ, refactor TerminalPshⱽ→ᴰ proofs for consistent style

maxsnew:
closes https://github.com/um-catlab/cubical-categorical-logic/issues/168

I implemented an Eq to PshIsoⱽ construction for convenience, and I did some minor refactorings like fixing the unicode triangles (\T not \t in emacs) and making a `Cubical.Data.Equality.More`

### DIFF
--- a/Cubical/Categories/Displayed/Presheaf/Constructions.agda
+++ b/Cubical/Categories/Displayed/Presheaf/Constructions.agda
@@ -6,7 +6,10 @@ open import Cubical.Foundations.Function
 open import Cubical.Foundations.HLevels
 open import Cubical.Foundations.Equiv.Base
 open import Cubical.Foundations.Equiv.Dependent
+
+import Cubical.Data.Equality as Eq
 open import Cubical.Data.Sigma
+open import Cubical.Data.Unit
 
 open import Cubical.Categories.Category
 open import Cubical.Categories.Bifunctor
@@ -33,6 +36,8 @@ open import Cubical.Categories.Displayed.Presheaf.Morphism
 
 open Bifunctorᴰ
 open Functorᴰ
+open isIsoOver
+open PshHomᴰ
 private
   variable
     ℓ ℓ' ℓᴰ ℓᴰ' : Level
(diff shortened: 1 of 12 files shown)


Cubical/Categories/Displayed/Presheaf/Constructions.agda
```diff
@@ -41,7 +46,62 @@ private
...
+  {P : Presheaf C ℓP} {Pᴰ : Presheafᴰ P Cᴰ ℓPᴰ}
+  {Q : Presheaf C ℓQ} {Qᴰ : Presheafᴰ Q Cᴰ ℓQᴰ}
+  where
+  private
+    module Pᴰ = PresheafᴰNotation Pᴰ
+  -- TODO: naming...
+  Lift-F-Homᴰ : ∀ {α : PshHom P Q} (αᴰ : PshHomᴰ α Pᴰ Qᴰ) → PshHomᴰ α (LiftPshᴰ Pᴰ ℓ) (LiftPshᴰ Qᴰ ℓ')
+  Lift-F-Homᴰ αᴰ .N-obᴰ pᴰ = lift (αᴰ .N-obᴰ (pᴰ .lower))
+  Lift-F-Homᴰ αᴰ .N-homᴰ {fᴰ = fᴰ}{pᴰ = pᴰ} i =
+    lift (αᴰ .N-homᴰ {fᴰ = fᴰ}{pᴰ = pᴰ .lower} i)
```
stschaef, line 89: Mathematically, we're treating `LiftPshᴰ` like a displayed endofunctor on presheaves. It would be nice if this were made explicit in the code, but this isn't possible. It would require a version of `PRESHEAFᴰ` where homs are `PshHomᴰ` rather than `NatTransᴰ, but the level polymorphism doesn't let this be realized as a displayed category

Do you think its worth making this sort of displayed functor construction explicit in the code? I ran into a similar situation yesterday where I wanted `reind` to be a displayed functor, but could not explicitly make it one for the same reason

stschaef (approved):
Just a single comment about how we may want to generalize some constructions, but otherwise LGTM

Cubical/Categories/Displayed/Presheaf/Constructions.agda
maxsnew: > Do you think its worth making this sort of displayed functor construction explicit in the code?

I think we should just do it as needed. But it's definitely a good thing to think about in terms of API design: if the thing you are doing looks like a level-polymorphic version of displayed category stuff, it's probably the right design.

PR #172, opened 2025-08-27
Two definitions of Exponential of Presheaves, Function types in SCwF

maxsnew:
This PR defines the exponential of presheaves in two ways:
1. The "large" version which is the usual definition using natural transformations
2. A "small" version which requires that the domain of the exponential be "locally representable", i..e, that the product with all representables is representable.

I'm not really sure what to call them but I chose large/small because the universe level of the small one is smaller. The large one is used to define the CCC structure on presheaves, whereas the small one is used to define exponentials in a category and function types in a SCwF. I haven't deleted any old definitions of Exponential yet though, I'll leave that for a big cleanup PR down the line.

Lastly, I don't know if we need it for anything, but I did prove that these two definitions are PshIso, which is a pretty faithful reproduction of the usual Yoneda-based argument.

### DIFF
--- a/Cubical/Categories/Displayed/Functor/More.agda
+++ b/Cubical/Categories/Displayed/Functor/More.agda
@@ -6,6 +6,7 @@ open import Cubical.Foundations.Prelude
 open import Cubical.Categories.Category.Base
 open import Cubical.Categories.Functor
 import      Cubical.Data.Equality as Eq
+import      Cubical.Data.Equality.More as Eq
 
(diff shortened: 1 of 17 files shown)


stschaef (approved):
This one also looks good

I have some amorphous thoughts on naming that, but they're not concrete enough to show up here.

For instance, I don't love the small/large split. And think that we should maybe refactor some of the library to make use of the terms local/global, but I don' think that fits well into this PR and can maybe come later in a renaming PR

PR #175, opened 2025-08-29
Implement displayed and vertical versions of ⇒PshSmall + supporting lemmas

maxsnew:
The displayed version is straightforward, but the vertical one was more difficult because the regular `Notation`s weren't powerful enough to handle it, because the `F-id` and `F-seq` case involve a path on objects (because we are using a morphism to take a cartesian lift). I ended up using several new combinators about reasoning about paths which were very useful and should probably be upstreamed sooner rather than later.

We should consider making more general versions of the `Presheaf` and `HomReasoning` modulues so that this doesn't all have to be done so manually, but I'll leave that for a future PR.

I also split `Displayed.Presheaf.Constructions` up into two modules because it was getting very long to re-typecheck.

closes https://github.com/um-catlab/cubical-categorical-logic/issues/162


stschaef (changes requested):
To close #162 I think this PR should also include a definition of `Exponentialⱽ` using these new definitions

I just pushed one if you'd like to look it over

maxsnew:
I had one already done from earlier this morning so I just pushed that

stschaef (approved):
perfect

PR #185, opened 2025-09-04
Total category of a displayed cartesian category is cartesian

hejohns:
Using what I understand @maxsnew to have talked about, about avoiding manually constructing universal elements by showing that the two presheaves are isomorphic


maxsnew:
We already had that PshIso implemented for something to do with displayed exponentials, so I refactored it to use that.

PR #200, opened 2025-09-23
Helpers from Progress on Reindexing Quantifiers

stschaef:
This PR defines `reindHet∀PshIsoⱽ`, which will be the core of the proof that a quantifier can be reindexed. However, after defining the UMP for quantifiers, it is likely this proof is refactor. Further, there are a couple comments about parts of the proof that are currently module parameters but should likely be upgraded to concrete constructions

In particular, `Fᴰ-weakening-NatTransᴰ` should be definable, which would change the module parameter `Fᴰ-NatIsoᴰ` to instead just be the property that `Fᴰ-weakening-NatTransᴰ` is an iso. However, when trying to handle the `N-hom` part of `Fᴰ-weakening-NatTransᴰ`, I am not able to use combinators that I would otherwise expect to be able to apply. For instance, it appears like a `sym reind-filler` should apply in a subgoal it does not. Apologies that this paragraph is deliberately rather rushed and unclear. I'm not giving too much detail here, because that is still being worked on and this PR is partial progress of helpers and such

This PR also includes additions to the displayed natural transformation library and some high level module reshuffling (like getting rid of `Displayed.NaturalIsomorphism` and moved into `Displayed.NaturalTransformation.More`)

There is also some bits that I thought would be helpful, but have not yet helped greatly: profunctor heteromorphisms and preservation of displayed universal elements. Each of these additions are no worse than whats already on main, but they did not help me as I'd expected them, so they're liable to bitrot


Cubical/Categories/Displayed/Presheaf/Constructions/Quantifiers/Base.agda
```diff
@@ -0,0 +1,235 @@
...
+  variable
+    ℓC ℓC' ℓCᴰ ℓCᴰ' ℓ ℓ' ℓP ℓPᴰ ℓQ ℓQᴰ ℓD ℓD' ℓDᴰ ℓDᴰ' : Level
+
+open NatTrans
+open Functor
+open Functorᴰ
+open PshHomᴰ
+open UniversalElementⱽ
+
+-- The universal/pi and existential/weak sigma type are defined as
```
maxsnew, line 54: This exact comment is also in Displayed/Quantifiers/Base

Cubical/Categories/Displayed/Presheaf/Constructions/Reindex/Properties.agda
```diff
@@ -241,6 +232,35 @@ module _ {C : Category ℓC ℓC'}{Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'}
...
+  {Pᴰ : Presheafᴰ P Cᴰ ℓPᴰ}
+  {Qᴰ : Presheafᴰ Q Cᴰ ℓQᴰ}
+  (αᴰ : PshIsoᴰ α Pᴰ Qᴰ)
+  where
+
+  private
+    module Pᴰ = PresheafᴰNotation Pᴰ
+    module Qᴰ = PresheafᴰNotation Qᴰ
+
+  PshIsoᴰ→PshIsoⱽ : PshIsoⱽ Pᴰ (reind (α .trans) Qᴰ)
```
maxsnew, line 247: This should not be implemented manually. This can be defined using `reind-introⱽ` and `reind-π`.

Cubical/Categories/Displayed/Presheaf/Morphism.agda
```diff
@@ -164,6 +181,16 @@ module _ {C : Category ℓC ℓC'} {Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'}
...
       isPshEqv
 
+module _ {C : Category ℓC ℓC'} {Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'}
+  {P : Presheaf C ℓP}
+  {Q : Presheaf C ℓQ}
+  {α : PshIso P Q}
+  {Pᴰ : Presheafᴰ P Cᴰ ℓPᴰ}
+  {Qᴰ : Presheafᴰ Q Cᴰ ℓQᴰ} where
+
+  pshIsoOfPshIsoᴰ : PshIsoᴰ  α Pᴰ Qᴰ → PshIso P Q
```
maxsnew, line 191: what is the point of this?

Cubical/Categories/FunctorComprehension/Base.agda
```diff
@@ -0,0 +1,71 @@
+{-# OPTIONS --lossy-unification #-}
```
maxsnew, line 1: Why did you move this file?

Cubical/Categories/Limits/BinProduct/More.agda
```diff
@@ -17,6 +21,7 @@ open import Cubical.Categories.Functor
 open import Cubical.Categories.FunctorComprehension
 open import Cubical.Categories.NaturalTransformation
 open import Cubical.Categories.Profunctor.General
+open import Cubical.Categories.Instances.Sets
```
maxsnew, line 24: why did you add these imports to this file?

Cubical/Categories/Profunctor/Homomorphism/Alt.agda
```diff
@@ -0,0 +1,72 @@
+module Cubical.Categories.Profunctor.Homomorphism.Alt where
```
maxsnew, line 1: Is this Profunctor Homomorphism stuff used somewhere? If not I don't want to add it

maxsnew (changes requested):
Left some comments

Cubical/Categories/FunctorComprehension/Base.agda
stschaef: I had a sketch of a `FunctorComprehension.Properties` file that I wanted a directory for

In it, I'd hoped to prove an analogue of

```
 prodStrNatIso : NT.NatIso (F ∘F -×c.×aF) (-×Fc.×aF ∘F F)
 ```
 
 but where the product functor is replaced with a general functor comprehension.
 
 I deleted the properties file bc I did not finish that proof, but I was able to prove it concretely for binary products

Cubical/Categories/Profunctor/Homomorphism/Alt.agda
stschaef: I believe these are needed for the lemma I aimed to prove in `FunctorComprehension.Properties`, but they are now not used anywhere so I'll remove these too

PR #205, opened 2025-10-27
Universal Property of the Universal Quantifier of Presheaves

stschaef:
This PR can be summarized as establishing the following universal property
```
∀Psh-UMPᴰ :
  Iso
    (PshHomᴰ (mkProdPshHom Pⱽ α)
      (reind (π₁ Q (C [-, a ])) Qᴰ) Pⱽ)
    (PshHomᴰ α Qᴰ (∀ⱽPsh Pⱽ))
```

`α` a morphism of presheaves from `Q` to `よ Γ`
`Qᴰ` a displayed presheaf over `Q`
`Pⱽ` a displayed presheaf over `よ (Γ × a)`

A displayed morphism of presheaves from `Qᴰ` into the universal quantifier of `Pⱽ` is the same data as a displayed morphism  _into_ `Pⱽ` out of a reindexed `Qᴰ` (where it is reindexed to be over `Q × よ a`)

Because the universal quantifier of presheaves is defined to be the reindexing of composition with the weakening functor, the universal property decomposes into two pieces. 

First I prove a universal property of just composition with the weakening functor in `Displayed.Presheaf.Constructions.UniversalProperty.ComposeWeakening`.

Then I combine this with the universal property of reindexing and give the full universal property for quantifiers in `Displayed.Presheaf.Constructions.UniversalProperty.Quantifiers`

### DIFF
--- a/Cubical/Categories/Displayed/Presheaf/Constructions/Quantifiers/Base.agda
+++ b/Cubical/Categories/Displayed/Presheaf/Constructions/Quantifiers/Base.agda
@@ -383,8 +383,5 @@ module _
             ∙ Pⱽ.≡in (αᴰ .N-homᴰ)
             ∙ Pⱽ.⟨⟩⋆⟨ Pⱽ.reind-filler _ _ ⟩
 
-        ∀ⱽPsh-introᴰ :
-          PshHomᴰ α
-              Qᴰ
-              ∀ⱽPsh
+        ∀ⱽPsh-introᴰ : PshHomᴰ α Qᴰ ∀ⱽPsh
         ∀ⱽPsh-introᴰ = reind-introᴰ ∀ⱽPsh-introᴰ'


Cubical/Categories/Displayed/Functor/More.agda
```diff
@@ -171,3 +172,41 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'} {F : Functor C D}
       (Fᴰ .F-homᴰ)
       (Gᴰ .F-homᴰ)
 
+module FunctorᴰNotation {C : Category ℓC ℓC'} {D : Category ℓD ℓD'} {F : Functor C D}
```
maxsnew: I think the most sensible thing in light of the below comments would be to just `open Functor (∫F F) public`. Or just use `∫F F` directly, though that's maybe less readable.

PR #208, opened 2025-11-03
Uncurried Variant of Displayed Presheaves

maxsnew:
This adds a new, equivalent definition of a displayed presheaf, which I call an "uncurried" displayed presheaf. See `Cubical.Categories.Displayed.Presheaf.Uncurried.Base` for an explanation. I also added the equivalence with our curried displayed presheaves and started reformulating our vertical universal properties to use the uncurried version, which generally works much more cleanly. The general pattern seems to be that displayed universal properties work well with curried presheaves but vertical universal properties work well with uncurried presheaves.

As an example of why uncurried presheaves make things easier, see `Cubical.Categories.Displayed.Presheaf.Uncurried.Constructions`, where I was able to define all of our main constructions on displayed presheaves (unit, lift, product, reindexing, and *large* versions of exponential and forall) without any manual definitions, all just using combinators on ordinary presheaves. I haven't gotten to the small versions yet but I have included comments that explain how they should work.

I've also initiated porting the gluing code to use uncurried vertical presheaves. So far I just have that fibrations (with the new definition) are preserved by reindexing.


Cubical/Categories/Displayed/Presheaf/Uncurried/Base.agda
```diff
@@ -0,0 +1,203 @@
...
+  where
+
+  _/Fⱽ_ : (Fᴰ : Functorⱽ Cᴰ Dᴰ) → (α : PshHom P Q) → Functor (Cᴰ / P) (Dᴰ / Q)
+  Fᴰ /Fⱽ α = Fᴰ /Fᴰ (α ⋆PshHom reindPshId≅ Q .trans)
+
+-- Interestingly, this one is at a lower universe level than Presheafᴰ
+
+-- Undecided about the naming here. ' is pretty bad but will do as a
+-- temporary solution. Should we just call it Presheafᴰ and use
+-- modules to resolve the difference?
```
stschaef: To the question here, I think yes. I think its better to gives the sane local name that you want and then bridge across these local names with modules/renaming rather than accumlating a bunch of `'` debt

Cubical/Categories/Displayed/Presheaf/Uncurried/Base.agda
```diff
@@ -0,0 +1,203 @@
...
+    ℓC ℓC' ℓCᴰ ℓCᴰ' : Level
+    ℓD ℓD' ℓDᴰ ℓDᴰ' : Level
+    ℓP ℓQ ℓR ℓPᴰ ℓPᴰ' ℓQᴰ ℓQᴰ' ℓRᴰ : Level
+
+open Category
+open Functor
+open Functorᴰ
+open PshIso
+
+-- TODO: better name...
```
stschaef: The best alternative I can come up with is `Cᴰ ×∫ P`, which is confusing in its own ways.

Maybe ``Cᴰ ×Elt P`? I'm not sure, but I agree that the slice notation isn't very intuitive. I think I see what you're getting at with it though. If `P` is representable, then its category of elements is a slice category. But even then, its still a confusing name after taking product with `Cᴰ`

Cubical/Categories/NaturalTransformation/More.agda
```diff
@@ -129,6 +129,10 @@ module _
   infixr 9 _⋆NatTrans_
   infixr 9 _⋆NatIso_
 
+  -- eqToNatIso : {F G : Functor C D}
```
stschaef: Changes to this file can be reverted

Cubical/Categories/Presheaf/Constructions/Reindex.agda
```diff
@@ -65,10 +69,33 @@ module _ {C : Category ℓC ℓC'} {D : Category ℓD ℓD'} where
...
-  reindPshIso F α .PshIso.trans = reindPshHom F (α .PshIso.trans)
-  reindPshIso F α .PshIso.nIso x .fst = α .PshIso.nIso _ .fst
-  reindPshIso F α .PshIso.nIso x .snd .fst = α .PshIso.nIso _ .snd .fst
-  reindPshIso F α .PshIso.nIso x .snd .snd = α .PshIso.nIso _ .snd .snd
+  reindPshIso F α .trans = reindPshHom F (α .trans)
+  reindPshIso F α .nIso x .fst = α .nIso _ .fst
+  reindPshIso F α .nIso x .snd .fst = α .nIso _ .snd .fst
+  reindPshIso F α .nIso x .snd .snd = α .nIso _ .snd .snd
+
+  -- this is right-whiskering
```
stschaef: With a better (locally small) category of presheaves, could this follow from a general whiskering principle?

I suspect yes, but I need to think about if size issues are relevant. That is, is the correct analogue here a whiskering by a `SmallFibNatTrans` or by a large natural transformation?

maxsnew: yea the only reason this isn't just whiskering is because it's PshHom instead of NatTrans

maxsnew:
Ok I did as suggested and removed the primes. The only one I kept was the file name Cartesian^V', which I only did because I think we will just get rid of the file Cartesian^V eventually anyway, so it's just a placeholder.

PR #209, opened 2025-11-03
Add flake.nix

stschaef:
@cxandru Is this akin to what you had in mind?

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ Everything.agda
 *.prof
 /.DS_Store
 /output/
+
+# Nix
+result
+.direnv/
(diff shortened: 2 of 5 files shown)


stschaef:
@maxsnew This should be good to merge. I've set up things so that
versioning of both Agda and `cubical` are both handled by the
flake. For Nixless users, I don't think this file interferes with anything.

I know that you've complained about managing multiple Agda versions,
and this makes it easier (but requires usage of `nix`).

This adds a Nix flake that pins exact versions of Agda and
the cubical library for the project. Instead of manually
installing the right versions system-wide, one can use `nix develop`
from the project directory to get a dev-shell with the appropriate
dependencies. 

I've also set this up locally to use `direnv` and `envrc.el` so that
when I enter the project directory, the desired dependencies are
automatically loaded. Likewise for Emacs buffers in this project. This
means that different projects could in theory have different Agda
versions, and there is no change in how you have to interact with
Emacs in order to have per-project configurations.

If you need to develop against a local `cubical` build you can override the
  input, `nix develop --override-input cubical path:../cubical`. We
  can also pin to precise versions by setting the commit hashes in the
  flake, or resync to using the latest builds by running `nix flake update`.
  
I would also wish to at some point in the future to refactor the CI/CD
to use Nix, but I think that's overkill for now.

PR #211, opened 2025-11-07
Locally small categories and globally small presheaves

stschaef:
This PR generalizes the notion of category by defining locally small categories, a category whose objects can be a large type but whose hom sets are all small

### DIFF
--- a/Cubical/Categories/LocallySmall/Displayed/Section/Base.agda
+++ b/Cubical/Categories/LocallySmall/Displayed/Section/Base.agda
@@ -115,11 +115,11 @@ module _ {C : Category Cob CHom-ℓ}
         F-homᴰ f Cᴰ.∫≡ F-homᴰ g
     F-homᴰ⟨ f≡g ⟩ i = (F.F-hom (f≡g i)) , (F-homᴰ (f≡g i))
 
-    ∫S : Functor D Cᴰ.∫C
-    ∫S .F-ob d = F.F-ob d , F-obᴰ d
-    ∫S .F-hom f = F.F-hom f , F-homᴰ f
-    ∫S .F-id = F-idᴰ
-    ∫S .F-seq f g = F-seqᴰ f g
+    intro : Functor D Cᴰ.∫C
+    intro .F-ob d = F.F-ob d , F-obᴰ d
+    intro .F-hom f = F.F-hom f , F-homᴰ f
+    intro .F-id = F-idᴰ
+    intro .F-seq f g = F-seqᴰ f g
 
 {-
    A Global Section is a local section over the identity functor.
--- a/Cubical/Categories/LocallySmall/Functor/Base.agda
+++ b/Cubical/Categories/LocallySmall/Functor/Base.agda
@@ -105,8 +105,8 @@ module _
      module Cᴰ = Categoryᴰ Cᴰ
      module ∫Cᴰ = Category Cᴰ.∫C
 
-   in[_]∫ : (c : Cob) → Functor Cᴰ.v[ c ] Cᴰ.∫C
-   in[ c ]∫ .F-ob = λ z → c , z
-   in[ c ]∫ .F-hom = λ z → Category.id C , z
-   in[ c ]∫ .F-id = refl
-   in[ c ]∫ .F-seq f g = sym $ Cᴰ.reind-filler _ _
+   σ : (c : Cob) → Functor Cᴰ.v[ c ] Cᴰ.∫C
+   σ c .F-ob = λ z → c , z
+   σ c .F-hom = λ z → Category.id C , z
+   σ c .F-id = refl
+   σ c .F-seq f g = sym $ Cᴰ.reind-filler _ _
--- a/Cubical/Categories/LocallySmall/Presheaf/GloballySmall/Base.agda
+++ b/Cubical/Categories/LocallySmall/Presheaf/GloballySmall/Base.agda
@@ -176,7 +176,7 @@ module _
       (λ _ → ΣPathP (refl , funExt λ _ → sym $ C.⋆Assoc _ _ _ ))
 
   YONEDA : Functor C.cat (∫C (PRESHEAF C))
-  YONEDA = ∫S YONEDA-S
+  YONEDA = intro YONEDA-S
 
 module _ {C : SmallCategory ℓC ℓC'} where
   private


stschaef:
# Utilities
Moved `Σω` stuff to `Data.Sigma.More`. Added a couple `isSet`-ish reasoning principles to `Cubical.Foundations`. Accidentally pushed a change to the `.agda-lib` file which I will revert.

# `LocallySmall`

I try to use module names that mirror the existing structure of the modules in `cubical`/`cubical-categorical-logic`. This PR should be contained in the directory `Cubical.Categories.LocallySmall` (It currently isn't so I will patch that up).

`LocallySmall` is meant to be analagous to `Cubical.Categories` and `LocallySmall.Displayed` is meant to be analagous to `Cubical.Categories.Displayed`.

## `LocallySmall.Category`

Definition of a locally small category in `Base`. `Small` provides several notions of smallness. The most widely used on is the `SmallCategory` record which is equivalent to the existing notion of `Category` from `Cubical.Categories.Category.Base` (upstream).

## `LocallySmall.Constructions`

### `BinProduct`
We give the product of locally small categories (`_×C_`) and the product of a locally small category by a small category (`_⊘_`).

#### `BinProduct.Properties`
This file provides an intro principle for functors into a binary product (`_,F_`). It similarly provides `,F-SFFunctorⱽ`. I'll expand on it more later, but this is both a bad name and likely shouldn't be in this file anyway.

## `LocallySmall.Functor`
Defines functors between locally small categories. This is pretty barebones. I also give a defintion of constant functor for locally small categories


## `LocallySmall.Bifunctor`
Bifunctors between locally small categories. This file also includes displayed bifunctors and should be separated.

## `LocallySmall.NaturalTransformation`
Provides
- `LargeNatTrans` as a general type of natural transformations between functors of locally small categories. These live at `Typeω` and are seemingly too big to be useful
- `NatTrans` as a type of natural transformations between functors from a small category into a globally small category. The restrictions on sizes here guarantee that the results type of natural transformations is small. However, this notion of natural transformation is too restrictive to be useful for reasoning about presheaves. Because these natural transformations are small, they allow us to organize functors of this kind into a category with these as morphisms. This category is given as `FUNCTOR` in the `NaturalTransformation` file
- `SmallFibNatTrans`. Given a small category `C` and a `SmallFibersCategoryᴰ` `Dᴰ` over a locally small category `D`, we can describe a type of natural transformations between functors `F : C → Dᴰ.v[ d ]` and `G : C → Dᴰ.v[ d' ]` along a morphism `d → d'`. This is the most useful notion of natural transformation as it is what allows us to describe the desired level polymorphism for categories of sets/presheaves. I further give the identity and composition rules for `SmallFibNatTrans` and equational reasoning principles, but I include the category for which these are the morphisms in `LocallySmall.Instances.Functor`.

This `NaturalTransformation` file is rather unorganized and inconsistent. These differing notions of natural transformations should be in their own file. Further, I provide the category of functors for `NatTrans` here but the category of functors for `SmallFibNatTrans` elsewhere, which is inconsistent. I also missed some comments that ought to be cleaned up.

## `LocallySmall.Instances`
Define terminal category, category of levels, indiscrete categories, and the small fibers functor category mentioned above.

## `LocallySmall.Displayed.Category`

Analgous to `LocallySmall.Category`. Give base defintions in `Base` and notions of smallness in `Small`.

## `LocallySmall.Displayed.Instances.Sets`
Gives the category of sets displayed over levels as a `SmallFibersCategoryᴰ`. I am conflicted where this should live. Mentally, I want to treat this like a category rather than a dispalyed one because we are hoping to generalize the ordinary category of sets. But techincally, this is a category displayed over levels.


## `LocallySmall.Displayed.Section.Base`
Tried to port `Categories.Displayed.Section` as faithfully as possible.

## `LocallySmall.Displayed.Section.Bisection`
Defined a notion of section for bifunctors. I thought this might be useful for a describing the binary product of presheaves, but that may be a nonstarter. Even if we wanted to base constructions on a notion of sections for bifunctors, I'm not sure if its nicer to factor things through the notions of sections in `Base` or if its worth defining the bifunctor sections as a primitive


## `LocallySmall.Displayed.Constructions`
Define weakening, reindexing, and displayed total category.

## `LocallySmall.Presheaf`
Define presheaves on a small category as the functor category arising from the `SmallFibNatTrans` notion of natural transformation.

Give definitions of the unit presheaf and the binary product of presheaves. Ports over presheaf notation and the yoneda embedding

In `Presheaf.Base` around line 80 I have an instance where I need to reason about a transport in `SET` but using `SET.reind-filler` doesn't have good inference of implicits so I have to use `transportRefl` directly. I'd like to avoid this if possible. The same snag for reasoning in `SET` appears in the definition of `×PshIntro` in `LocallySmall.Presheaf.Constructions.BinProduct`

Cubical/Categories/LocallySmall/Presheaf/Base.agda
```diff
@@ -0,0 +1,216 @@
...
+
+open SmallFibNatTrans
+module _
+  {C : SmallCategory ℓC ℓC'}
+  where
+  private
+    module C = SmallCategory C
+  open Categoryᴰ
+
+  よ : Functor  C.cat (∫C (PRESHEAF C))
```
maxsnew: This terminology is inconsistent. In the existing library we call this YO in `Categories.Yoneda` or `YONEDA` in `Categories.Yoneda.More`

Cubical/Categories/LocallySmall/Presheaf/Base.agda
```diff
@@ -0,0 +1,216 @@
...
+      (λ _ → ΣPathP (refl , funExt λ _ → C.⋆IdR _))
+  よ .F-seq f g =
+    makeSFNatTransPath refl
+      (λ _ → ΣPathP (refl , funExt λ _ → sym $ C.⋆Assoc _ _ _ ))
+
+  HomLevelF : Functor C.cat LEVEL
+  HomLevelF = Constant (liftω ℓC')
+
+  open Section
+  よS : Section HomLevelF (PRESHEAF C)
```
maxsnew: Can you define よ above using this one instead of defining よ manually?

Cubical/Categories/LocallySmall/Displayed/Constructions/BinProduct/Base.agda
```diff
@@ -0,0 +1,168 @@
...
+  {D : Category Dob DHom-ℓ} {Dobᴰ-ℓ Dobᴰ DHom-ℓᴰ}
+  (Dᴰ : SmallFibersCategoryᴰ D Dobᴰ-ℓ Dobᴰ DHom-ℓᴰ)
+  where
+  private
+    module C = CategoryNotation C
+    module Cᴰ = Categoryᴰ Cᴰ
+    module D = CategoryNotation D
+    module Dᴰ = CategoryᴰNotation Dᴰ
+
+  _×CᴰSF_ : SmallFibersCategoryᴰ (C ×C D) _
```
maxsnew: Why isn't this just ×Cᴰ ?

Cubical/Categories/LocallySmall/Constructions/BinProduct/Base.agda
```diff
@@ -0,0 +1,91 @@
...
+  π₂ .F-id = refl
+  π₂ .F-seq _ _ = refl
+
+module _
+  (C : SmallCategory ℓC ℓC')
+  (D : Category Dob DHom-ℓ) where
+  private
+    module C = SmallCategory C
+    module D = CategoryNotation D
+  _⊘_ : Category (C.small-ob ×ω Dob) _
```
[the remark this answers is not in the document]
stschaef: Similar to the below comment about `_×CᴰSF_`, the definitions of objects here may differ slightly at `Typeω`.
`C.small-ob ×ω Dob` (used for `⊘`) and `Σω[ c ∈ Cob ] Dob` (`×C`) aren't on the nose the same. 

Initially I thought that it might be wise to preserve the definition of `_⊘_` in case these differences end up being relevant. After some thought, I can't think of a situation where I'd want `_⊘_` so I've tentatively removed it in my most recent push

maxsnew (changes requested):
left a couple more comments

PR #219, opened 2025-11-19
Small Forall, Exponential of Displayed Presheaves, Existential quantifier

maxsnew:
The main purpose of this pull request is to get a definition of small
vertical exponential and universal quantifier that are isomorphic to
the large versions. This is done, and we now are back to having a
definition of vertical CCC.

The main theorems I needed to prove are that the local
representability-type conditions we had identified for them before
imply the representability condition required by the generic right
adjoint (Presheaf.Constructions.RightAdjoint) interface. Those
theorems are `LR∀-repr` and
`LocallyRepresentableⱽ→LocallyRepresentable`. These are non-trivial to
establish because the generic right adjoint interface is asking for
things to be representable in the slice category `Cᴰ / P` rather than
to be Representableⱽ/ᴰ in Cᴰ, like we are used to dealing with.

The key insight is that every representable in the slice category is
equivalent to a *pushforward* of a displayed representable (see
`push-repr`). The pushforward construction is the left adjoint to the
pullback/reindexing along a natural transformation `push`. I define it
from some other new primitives: existential/weak sigma and a "path
presheaf".

Then, coming back to the vertical exponentials and the universal
quantifiers, each one requires proving a technical lemma involving the
pushforward. For the vertical exponentials, we need Frobenius
reciprocity (`FrobeniusReciprocity`), and for universal quantifiers we
need the Beck-Chevalley (`BeckChevalley`). For Frobenius I ended up
defining a general version of it for arbitrary Types, which I couldn't
find upstream. Formalizing Beck Chevalley also required giving a
definition of when a square of PshHoms is a pullback.

Other than that, the PR is mostly cleanup of a few messy proofs to use
nicer lemmas (`F⇒Small≅F⇒Large`, `-×Psh_-cocontinuous`,
`reindPshF-cocont`). This includes defining the right notion of
fully-faithfulness for displayed functors and proving that the
projection out of a reindexed displayed category is fully faithful.

I left in a couple of commented out parts: some in RightAdjoint that
need some universe polymorphic PshHom and another in CartesianClosedV
where I encountered a huge performance degradation from using records
rather than sigma types, which I just wanted to document.

### DIFF
--- a/Cubical/Categories/Displayed/Presheaf/Uncurried/Constructions.agda
+++ b/Cubical/Categories/Displayed/Presheaf/Uncurried/Constructions.agda
@@ -207,7 +207,6 @@ module _ {C : Category ℓC ℓC'}{Cᴰ : Categoryᴰ C ℓCᴰ ℓCᴰ'} where
     -- Pᴰ(p) ⊢ (Σ[ p' ] Pᴰ(p') × αp'≡αp)
     -- ===============================
     -- Pᴰ(p) ⊢ (Σ[ p' ] Pᴰ(p') × αp'≡αp)
-
     push-σ : PshHomᴰ α Pᴰ push
     push-σ .N-ob (Γ , Γᴰ , p) pᴰ = p , (pᴰ , refl)
     push-σ .N-hom _ _ (γ , Γᴰ , γ⋆p≡p') pᴰ = ΣPathP ((sym $ γ⋆p≡p') , (ΣPathPProp (λ _ → Q.isSetPsh _ _)
@@ -317,8 +316,6 @@ module _ {C : Category ℓC ℓC'} where
       test : ∀ x → PB.p[ x ] ≡ (Σ[ (p , q) ∈ P.p[ x ] × Q.p[ x ] ] α .N-ob x p ≡ β .N-ob x q)
       test x = refl
 
-    -- pb-π : PshHom Pullback (P ×Psh Q)
-
     module _ {S : Presheaf C ℓS}
       (α' : PshHom S Q) (β' : PshHom S P)
       where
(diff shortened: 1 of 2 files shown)


Cubical/Categories/Displayed/Presheaf/Uncurried/Constructions.agda
```diff
@@ -8,152 +17,438 @@ open import Cubical.Foundations.Isomorphism
...
+        λ αⱽ → makePshHomPath $ funExt λ (x , xᴰ , p) → funExt λ (q , pᴰ) → refl
+  module _ (P : Presheaf C ℓP) where
+    private
+      module P = PresheafNotation P
+    PathPsh' : Functor ((∫C (Element (P ×Psh P))) ^op) (PROP ℓP)
+    PathPsh' = mkFunctor (PROP _) hasPropHomsPROP (λ (_ , p , p') → (p ≡ p') , P.isSetPsh p p')
+      λ {(x , p , p')} {(y , q , q')} (f , fp,fp'≡q,q') p≡p' →
+        (sym $ PathPΣ fp,fp'≡q,q' .fst) ∙ cong (f P.⋆_) p≡p' ∙ PathPΣ fp,fp'≡q,q' .snd
+
+    PathPsh : Presheafᴰ (P ×Psh P) Cᴰ ℓP
```
stschaef: We briefly discussed yesterday the idea of a displayed category of paths over the product of small categories, and how that construction may be represented by this presheaf. I'm curious if we'd ever benefit from expanding on this connection

That is, once a connective is defined for presheaves, we've used the presheaf construction at the representable to define the connective in the general case. Would we want to do the same here to express a general internal equality type?

Cubical/Categories/Instances/Props.agda
```diff
@@ -0,0 +1,37 @@
...
+open import Cubical.Categories.Functor
+open import Cubical.Categories.NaturalTransformation
+open import Cubical.Categories.Instances.Thin
+open import Cubical.Categories.Instances.Sets
+
+private
+  variable ℓ : Level
+
+open Functor
+-- TODO: make a general "Thin category/Preorder→Category" construction
```
stschaef: Is this referring to whats in `Instances.Thin` or something distinct?

Cubical/Categories/Presheaf/Representable/More.agda
```diff
@@ -254,3 +259,23 @@ module _ {C : Category ℓc ℓc'}{P : Presheaf C ℓp}{Q : Presheaf C ℓq} whe
     _◁PshIso_ .vertex = ue.vertex
     _◁PshIso_ .element = α .trans .N-ob ue.vertex ue.element
     _◁PshIso_ .universal = seqIsUniversalPshIso ue.universal α
+
+
+module _ {C : Category ℓc ℓc'} where
+  □ : Functor (PresheafCategory C ℓp) (PresheafCategory C _)
```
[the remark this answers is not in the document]
stschaef: I think that makes some sense if I stop to think, but a comment would help

stschaef (commented):
Looks pretty good to me. Just a couple comments

Cubical/Categories/Instances/Props.agda
maxsnew: deleted

stschaef (approved):

PR #225, opened 2025-11-30
Cbpv small step

bond15:


maxsnew:
What's the status of this? Are we going to incorporate this or drop it?

PR #245, opened 2026-04-10
Alternative pullback definition

stschaef:
Adds `Cubical.Categories.Limits.Pullback.Alt` , an alternative
formulation of pullbacks as binary products in the slice
category. Also defines the base change functor and shows that sets have these kinds of pullbacks

As much as possible, I would prefer to bootstrap more complicated
constructions, like the pullback, with existing parts of the library

### DIFF
--- a/Cubical/Categories/Instances/Sets/Properties.agda
+++ b/Cubical/Categories/Instances/Sets/Properties.agda
@@ -6,6 +6,7 @@ open import Cubical.Foundations.Equiv
 open import Cubical.Foundations.HLevels
 open import Cubical.Foundations.Isomorphism
 open import Cubical.Foundations.Structure
+open import Cubical.Foundations.Function
 
 open import Cubical.Data.Sigma
 open import Cubical.Data.Sum as Sum
@@ -17,7 +18,7 @@ open import Cubical.HITs.SetCoequalizer as SetCoeq
 open import Cubical.Categories.Category
 open import Cubical.Categories.Bifunctor
 open import Cubical.Categories.Exponentials
-open import Cubical.Categories.Instances.Sets
+open import Cubical.Categories.Instances.Sets hiding (PullbacksSET)
 open import Cubical.Categories.Presheaf.Representable
 open import Cubical.Categories.Limits.Cartesian.Base
 open import Cubical.Categories.Limits.Coend
@@ -26,6 +27,7 @@ open import Cubical.Categories.Limits.BiCartesianClosed.Base
 open import Cubical.Categories.Limits.Terminal.More
 open import Cubical.Categories.Limits.BinProduct.More
 open import Cubical.Categories.Limits.IndexedProduct.Base
+open import Cubical.Categories.Limits.Pullback.Alt
 
 private
   variable ℓ ℓC ℓC' : Level
@@ -140,3 +142,23 @@ module _ {ℓSET : Level} where
   SETBiCCC .BiCartesianClosedCategory.CCC = SETCCC
   SETBiCCC .BiCartesianClosedCategory.sums = BinCoProductsSET
   SETBiCCC .BiCartesianClosedCategory.init = InitialSET
+
+module _ {ℓSET : Level} where
+  PullbacksSET : Pullbacks (SET ℓSET)
+  PullbacksSET {l = l}{r = r} f g .vertex .fst .fst =
+    Σ[ (x , y) ∈ ⟨ l ⟩ × ⟨ r ⟩ ] f x ≡ g y
+  PullbacksSET {l = l}{m = m}{r = r} f g .vertex .fst .snd =
+    isSetΣ (isSet× (l .snd) (r .snd)) λ _ → isProp→isSet (m .snd _ _)
+  PullbacksSET f g .vertex .snd ((x , y) , e) = f x
+  PullbacksSET f g .element .fst .fst ((x , y) , e) = x
+  PullbacksSET f g .element .fst .snd = refl
+  PullbacksSET f g .element .snd .fst ((x , y) , e) = y
+  PullbacksSET f g .element .snd .snd = funExt λ ((x , y) , e) → sym e
+  PullbacksSET {m = m} f g .universal (u , h) =
+    isIsoToIsEquiv $
+      (λ x → (λ z → (x .fst .fst z , x .snd .fst z) ,
+                    funExt⁻ (x .fst .snd ∙ (sym (x .snd .snd))) z) , (x .fst .snd)) ,
+      (λ _ → ΣPathP ((ΣPathPProp (λ _ → isSet→ (m .snd) _ _) refl) ,
+                    (ΣPathPProp (λ _ → isSet→ (m .snd) _ _) refl))) ,
+      (λ _ → ΣPathP ((funExt λ _ → ΣPathPProp (λ _ → m .snd _ _) refl) ,
+                    isProp→PathP (λ _ → isSet→ (m .snd) _ _) _ _))
--- a/Cubical/Categories/Instances/Sets/Properties.agda
+++ b/Cubical/Categories/Instances/Sets/Properties.agda
@@ -12,6 +12,8 @@ open import Cubical.Data.Sigma
 open import Cubical.Data.Sum as Sum
 open import Cubical.Data.Empty as Empty
 open import Cubical.Data.Unit
+import Cubical.Data.Equality as Eq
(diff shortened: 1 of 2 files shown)


stschaef:
I am still often a git noob, so sanity check this.

This PR is the start of a chain of dependencies between PRs. In order of how they should be merged,
#245 < #246 < #247 < #248 < #249 < #250

stschaef:
I have each successive PR based at the previous one in the chain

Cubical/Categories/Limits/Pullback/Alt.agda
```diff
@@ -0,0 +1,213 @@
...
+open PshHom
+open PshIso
+open NatTrans
+
+module _ (C : Category ℓ ℓ') where
+  private
+    module C = Category C
+
+  module _ (m : C.ob) where
+    Slice : Categoryᴰ C _ _
```
maxsnew: This is not the slice category, the total category of this category is the slice category.

Cubical/Categories/Limits/Pullback/Alt.agda
```diff
@@ -0,0 +1,213 @@
+{-# OPTIONS --lossy-unification #-}
+-- A definition of pullbacks bootstapped using
+-- products in the total category of slices
```
maxsnew: A better comment here would be to describe the construction in the file and also list the major theorems (which here is apparently ChangeBase)

Cubical/Categories/Limits/Pullback/Alt.agda
```diff
@@ -0,0 +1,213 @@
...
+  module PullbacksNotation (pbs : Pullbacks C) where
+    module pb {l m r} {f : C [ l , m ]} {g : C [ r , m ]} =
+      PullbackNotation (pbs f g)
+    open pb public
+
+module _ {ℓC ℓC'} (C : Category ℓC ℓC') (pb : Pullbacks C) {x y} (f : C [ y , x ]) where
+  private
+    module C = Category C
+    module pb {z} {g} = PullbackNotation {r = z} {f = f} {g = g} (pb f g)
+  ChangeBase : Functor (∫C (Slice C x)) (∫C (Slice C y))
```
maxsnew: This functor is right adjoint to a post-composition functor f : Slice y -> Slice x, so you can use functor comprehension to handle these proofs.

Cubical/Categories/Limits/Pullback/Alt.agda
```diff
@@ -0,0 +1,213 @@
...
+    module C = Category C
+
+  module _ (m : C.ob) where
+    Slice : Categoryᴰ C _ _
+    Slice = Element (C [-, m ])
+
+  module _ {m : C.ob} where
+    module _ {l r} (f : C [ l , m ]) (g : C [ r , m ]) where
+      Pullback : Type _
+      Pullback = BinProduct (∫C (Slice m)) ((l , f) , (r , g))
```
maxsnew: The main one I am thinking of is that for any category `C` we can define a displayed category `Slices C` over `C` whose objects over `c` are `d` and `f : d -> c` (this can be defined as the simple total category of the displayed `Arrow` category). This displayed category is the displayed category version of the [codomain "fibration"](https://ncatlab.org/nlab/show/codomain%20fibration). `Slices C` is always an opfibration (given by composition). A cartesian lift of `g : x -> y` along `f : z -> y` in `Slices C` is equivalent to a pullback. This seems like a more natural definition to me (e.g., `ChangeBase` below is then the action of a cartesian lift on (something equivalent to) fiber categories).

maxsnew (changes requested):
Not sure if this is the best definition to base things off of. See my comment about using cartesian lifts instead.

Cubical/Categories/Limits/Pullback/Alt.agda
stschaef: Renamed this to `Sliceᴰ` and used the name `Slice` for its total category

Cubical/Categories/Limits/Pullback/Alt.agda
stschaef: Just pushed a version with this, and it's much cleaner this way

Cubical/Categories/Limits/Pullback/Alt.agda
stschaef: I agree that what you describe here would work, but I am split on which is better or more natural. Are there technical reasons why you would prefer one over the other?

stschaef: I like the natural reading of (op)cartesian lifts in your construction, but the asymmetry seems a little unnatural. Like, the morphism that you're lifting along and the arrow that you have upstairs are symmetric mathematically when describing a pullback, but they're occurring in different places here

Although other than aesthetics, I can't think of any downside from this presentation

maxsnew: In some settings one is more natural, in others the other one is. They are equivalent. The one where it is a cartesian lift just seems better suited to what you are using it for (change base)

Cubical/Categories/Limits/Pullback/Alt.agda
maxsnew: Does this terminology really make sense? `Sliceᴰ` sounds to me like I am taking a slice category in a displayed category, or a "slice category over a slice category"

Cubical/Categories/Instances/Sets/Properties.agda
```diff
@@ -140,3 +142,23 @@ module _ {ℓSET : Level} where
   SETBiCCC .BiCartesianClosedCategory.CCC = SETCCC
   SETBiCCC .BiCartesianClosedCategory.sums = BinCoProductsSET
   SETBiCCC .BiCartesianClosedCategory.init = InitialSET
+
+module _ {ℓSET : Level} where
+  PullbacksSET : Pullbacks (SET ℓSET)
+  PullbacksSET {l = l}{r = r} f g .vertex .fst .fst =
+    Σ[ (x , y) ∈ ⟨ l ⟩ × ⟨ r ⟩ ] f x ≡ g y
```
maxsnew, line 151: Eq?

Cubical/Categories/Limits/Pullback/Alt.agda
```diff
@@ -0,0 +1,282 @@
...
+--     pbPr₂ : C [ pbOb , cspn .r ]
+--     pbCommutes : pbPr₁ ⋆ cspn .s₁ ≡ pbPr₂ ⋆ cspn .s₂
+--     univProp : isPullback cspn pbPr₁ pbPr₂ pbCommutes
+--
+-- Likewise, the definition of the slice category found in
+-- Cubical.Categories.Instances.Slice defines manual record types
+-- to describe the objects and morphisms in the slice category
+--
+-- Although it is straightforward to define these data, it is
+-- verbose and overly concrete. Instead, we can provide a
```
maxsnew: I don't think you need to be so derisive towards the other definition. It's perfectly fine. You can just say this file provides an alternate, equivalent definition that is constructed compositionally from other notions

Cubical/Categories/Limits/Pullback/Alt.agda
stschaef: To avoid confusion, I removed `Sliceᴰ` altogether and just defined `Slice` to directly be the total cat

Cubical/Categories/Limits/Pullback/Alt.agda
stschaef: the most recent comment should be a bit more balanced

Cubical/Categories/Instances/Sets/Properties.agda
stschaef: I've provided `PullbacksSET'` that uses an `Eq` instead of a path. Although, I have to use `eqToPath` and `pathToEq` in combination with function extensionality for paths, so I'm curious if this will have any impacts on usability

maxsnew: yea I guess just get rid of this. You would really want this if you had a version of Pullback based on EqElement.

Cubical/Categories/Limits/Pullback/Alt.agda
maxsnew: You should also be able to get much less manual proofs of ChangeBaseId and ChangeBaseComp this way by showing the corresponding facts about their left adjoints (PostComposeWith Id = Id and similar for comoposition)

maxsnew:
@stschaef can you finish this before it bitrots

PR #247, opened 2026-04-10
Define strict functors

stschaef:
- Defines a type of functors such that the unit and associativity laws hold definitionally
- Defines a notion of natural transformation for these functors, but this type of natural transformations does not enjoy the same strictness properties. I don't think a type of general natural transformations ever can, because of `N-ob` being valued in an abstract type of morphism
- Also defines an accompanying notion of strict bifunctors and `PshHom` to interoperate with `StrictFunctor`. These are less thought out. I simply had Claude port a sufficient amount of code for using these strict functors when describing a double category of profunctors (to come in a later PR)

The motivating reason for `StrictFunctor` is that goals that mention a composition of functors `F ∘F Id` often force us to use heterogeneous paths because this composite is not definitionally `F`. We are able to use homogeneous paths by providing a type of functors for which this is a definitional equality. 

I'm curious but unsure if we would wish to use `StrictFunctor` as the default notion of functor. It relies on `eta-equality`, which isn't consistent with our recent `no-eta-equality` refactors

### DIFF
--- a/Cubical/Categories/Functors/Strict/Base.agda
+++ b/Cubical/Categories/Functors/Strict/Base.agda
@@ -1,0 +1,152 @@
+-- Functors that are definitonally unital and associative.
+-- Spiritually similar to the usage of PshHom over NatTrans
+--
+-- Should this be the default notion of functor?
+-- It does rely on eta-equality
+module Cubical.Categories.Functors.Strict.Base where
+
+open import Cubical.Foundations.Prelude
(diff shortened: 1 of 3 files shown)


stschaef:
In principle, this PR should probably only have the `StrictFunctor` stuff and not the rest of it. I tried to separate the chain of contributions into modular pieces, but I didn't fully decouple this PR from later pieces in the chain

PR #253, opened 2026-04-15
proof of concept for thickened displayed categories

maxsnew:
This is a quick and dirty sketch of how thickened displayed categories would work, up to the definition and isomorphism between curried and uncurried presheaves. I left an inordinate amount of holes but this at least shows that the approach we have so far to universal properties doesn't fundamentally have to change.

The real test would be defining vertical/displayed universal properties and seeing if e.g., reindexing theorems are easier


Cubical/Categories/Instances/ExtraId.agda
```diff
@@ -0,0 +1,151 @@
...
+
+  elimProp3 : ∀ {x}
+    → {M : ∀ y (f : ExtraIdHom x y) z (g : ExtraIdHom y z) w (h : ExtraIdHom z w) → Type ℓ}
+    → (∀ {y}{z}{w} f g h → isProp (M y f z g w h))
+    → (MHom : ∀ {y z w} f g h → M y (semHom f) z (semHom g) w (semHom h))
+    → ∀ y f z g w h → M y f z g w h
+  elimProp3 isPropM MHom = elimProp2 (λ f g → isPropΠ2 (λ x₁ → isPropM f g))
+    (λ f g → elimProp (λ {y = y₁} → isPropM (semHom f) (semHom g)) (λ {y = y₁} → MHom f g))
+
+  -- elim2 : ∀ {x}
```
stschaef, line 89: Is `elim2` unimplemented bc its not necessary for your test/demonstration or is there a derper reason impeding this definition?

stschaef: If `⋆ExtraId` were implemented with `elim2`, would `IdR` also hold by `refl`?

Cubical/Categories/Instances/ExtraId.agda
```diff
@@ -0,0 +1,151 @@
+{-
+  Given a category C, produces an equivalent category ExtraId C whose Hom sets have a new identity morphism freely adjoined
```
stschaef, line 2: I find myself asking the naive question "can this approach be replicated with composition?" in a way that gives nicer definitional equality on something like `⋆Assoc`

For a category `C`, you've in some sense given a category generated by `C` along with a freely adjoined identity + coherence data constraining the freely generated identity to act like the existing identity. There seems to be a generalization of your approach to here where we don't just look at a freely adjoined identity element, but perhaps even a free category generated by all formal composites generated by the morphisms in `C`

stschaef: I'm not quite sure yet how this might be used, so I'm not suggesting using it at this moment, but I wanted to mention it as a natural generalization

stschaef: I suppose what you have is very close to that anyway. the only difference might be a formal composition operator in the new hom type along with path constructors constraining the behavior of formal compositions.

Cubical/Categories/Displayed/Thick/Base.agda
```diff
@@ -0,0 +1,428 @@
+{-
+  Alternate definition of displayed categories.
+
+  Equivalent to a lax functor functor to a bicategory of Profunctors.
+
+  I've sketched out the major definitions that we need for gluing, culminating in curried and uncurried variants of displayed presheaves.
+
+  Currently I've left all of the proofs as holes.
+-}
+module Cubical.Categories.Displayed.Thick.Base where
```
stschaef, line 10: Do you imagine keeping this file and providing lemmas relating `Thick.Base` with `Thick.Slick`, or do you imagine just using `Thick.Slick`? I'd imagine we want the latter because its implementable with no new "primitives", so to speak

Cubical/Foundations/Level.agda
```diff
@@ -0,0 +1,8 @@
+module Cubical.Foundations.Level where
+
+open import Cubical.Foundations.Prelude
+
+_⊔_ : Level → Level → Level
```
stschaef, line 5: In `Foundations.More` at `L316` we have 
```
_⊔ℓ_ : Level → Level → Level
_⊔ℓ_ = ℓ-max
```

We should probably delete the one in `More`, as I think `Level` is a better location; and, at some point `More` should have its definitions separated into the appropriate locations anyways

Cubical/HITs/MappingCylinder/Base.agda
```diff
@@ -0,0 +1,43 @@
+module Cubical.HITs.MappingCylinder.Base where
+
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.Equiv
+open import Cubical.Foundations.HLevels
+open import Cubical.Foundations.Isomorphism
+
+data MappingCylinder {ℓ ℓ'}{A : Type ℓ}{B : Type ℓ'}(f : A → B) : Type (ℓ-max ℓ ℓ') where
```
stschaef, line 8: Is this something like `join` of `B` and the image of `f`?

stschaef (commented):
this is really cool! just have a few questions

Cubical/HITs/MappingCylinder/Base.agda
stschaef: or more precisely, I think the following is true `MappingCylinder f ≅ join (Σ[ b ∈ B ] (fiber f b)) B`

Cubical/Categories/Instances/ExtraId.agda
maxsnew: I just haven't finished `elim2` yet. There's no reason it can't be done.

> If `⋆ExtraId` were implemented with `elim2`, would `IdR` also hold by `refl`?

No.

Cubical/HITs/MappingCylinder/Base.agda
stschaef: oh no, I think it's more subtle than this. Feel free to ignore

Cubical/Categories/Displayed/Thick/Base.agda
maxsnew: Yeah I implemented all of this before I came up with the slick definition, I don't know if it's necessary anymore at all

Cubical/Foundations/Level.agda
maxsnew: we should also just upstream this, it's ridiculous that it's not in `cubical`

Cubical/HITs/MappingCylinder/Base.agda
maxsnew: No, a join joins everything from the two sides. But both MappingCylinder and Join are special cases of a Pushout so that's why they look similar

Cubical/Categories/Instances/ExtraId.agda
stschaef: moving this to zulip

PR #254, opened 2026-04-24
Fix flake issues + update CI to use nix with up-to-date workflows

### LINKED ISSUE #244: Update CI
We currently build the entire library using `Everything.hs` to generate a bunch `Everything.agda` files, but this isn't sufficient for testing the `Guarded` module in #242 because it uses the infective `--guarded` flag which we don't handle. Rewrite CI to use `--build-library` flag, in the same manner as the `cubical` repo

Also update versions of actions used in the Github CI workflow, we have actions that have deprecated since September
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

stschaef:
the previous version of the flake had an unnoticed build issue that would cause `nix build` to fail because of insufficient permissions for writing interface files to a `cubical-categorical-logic/_build` subdirectory in the nix store. To fix this, I instead reworked the flake to be closer to how `cubical`s flake

Additionally, this should resolve #244. I reworked the CI to use nix for dependency tracking, and I changed our method of testing the entire repo to instead use Agda's `--build-library` flag. This avoids having to deal with the `gen-everythings.hs` and the `Everything.agda` files for each subdirectory. 

This method still uses a single `Everything` entrypoint for generating the rendered html version of the code, but not when typechecking the project. This entrypoint is now generated by the `generate_everything` shell script, which is much simpler than the previous Haskell file


maxsnew:
@stschaef merge this if it's ready

PR #255, opened 2026-05-08
Yoneda strictification of a category

stschaef:
This defines a strictification operation on categories that replaces the morphisms in an arbitrary category with `PshHomStrict`s between Yoneda embeddings

PR #256, opened 2026-05-08
Yoneda strictification of a presheaf

stschaef:
This defines a strictification operation for a presheaf by replacing the elements of a presheaf `P` at `x` with a `PshHomStrict` between the Yoneda embedding on `x`

This should remain a draft until #255 is merged, then it should be rebased and merged. For readability, I am going to initially base this PR off of `strict`, but this should also be edited to instead be based at `main` when we go to merge


maxsnew:
@stschaef do we have any applications of this

maxsnew:
@stschaef do we have any applications of this

stschaef:
> @stschaef do we have any applications of this

no, not at the moment. I'll close this

PR #257, opened 2026-05-08
Yoneda Strictification of a Displayed Category

stschaef:
This defines a notion of `PshHomStrictᴰ` for uncurried displayed presheaves, which is then used to provide a Yoneda strictification operation on displayed categories

This should remain a draft until #255 is merged, then it should be rebased and merged. For readability, I am going to initially base this PR off of `strict`, but this should also be edited to instead be based at `main` when we go to merge


stschaef:
This also now defines a displayed full image construction and instantiates the displayed strictification operation as the displayed full image of the strict displayed yoneda embedding

PR #265, opened 2026-08-02
Cbpv multiplicative + state enrichment gluing/canonicity

maxsnew:
This PR contains 2 new syntaxes and canonicity proofs for unary CBPV:
1. The first is for multiplicative unary CBPV, i.e., adding U and F
2. The second extends this with an enrichment of the computations with a specific algebraic structure of boolean state.

For U/F, the ordinary and vertical versions are just (op)cartesian lifts, the displayed version is a new definition of "displayed cartesian lift over a cartesian lift". The reindexing here is just an instantiation of the pre-existing reindexing of (op)cartesian lifts.

For state enrichment, I first developed notions of (displayed) algebra for the theory of a boolean state, whose free models are given by the state monad. Then I defined a notion of state-enrichment for a CBPV model, which basically says the computations carry a state algebra and the composition is a homomorphism, in addition to displayed versions. This also needed a reindexing theorem for displayed state enrichment along an enriched functor (I don't think there's any vertical version of this). I didn't use any pre-existing notion of enriched category or functor, I just made this all bespoke for state algebras specifically.

For the gluing and canonicity I added syntax for the two types and used two models: the existing set-based one for pure multiplicative, and a new displayed state alg model for the enriched one. The most interesting thing that I discovered doing this is that the universal construction for vertical F types. The universal construction is you have a computation `M : C [ A , B ]` and you want an opcartesian lift `push M Aᴰ`. In implementing this concretely for the state algebra model, I found that the construction naturally decomposed into two pieces: first construct the opcartesian lift of `ret : C [ A , F A ]` , so you get `push ret Aᴰ` over `A` and then construct an opcartesian lift of `rec M : C [ F A , B ]` getting `push (rec M) (push ret Aᴰ)` over `B`. Because opcartesian lifts compose, this gets you a lift of `η ⋆ rec M ≡ M` so you get the result with a transport. The push of `ret` looks like a displayed version of the free algebra, whereas the push of `rec M` looks like an ordinary image of an algebra homomorphism.

End-to-end it's got a few transports, so I think this might be a good target for improving definitional behavior because it all compiles relatively quickly compared to the exponentials. On the other hand, it's just a fact that `η ⋆ rec M ≡ M` doesn't hold definitionally so it's not just a matter of better universal properties.


maxsnew:
sorry about that just pushed another small re-org

Cubical/Algebra/State.agda
```diff
@@ -0,0 +1,243 @@
+-- Algebras for the theory of a Boolean state.
+-- The free algebras are given by the state monad.
+{-# OPTIONS --prop #-}
+module Cubical.Algebra.State where
```
stschaef, line 3: should this PR use the general construction I poked at in `cbpv-mult-strict`?

Cubical/Algebra/State.agda
```diff
@@ -0,0 +1,243 @@
+-- Algebras for the theory of a Boolean state.
+-- The free algebras are given by the state monad.
+{-# OPTIONS --prop #-}
```
stschaef: This `--prop` is not neceesary

Cubical/Categories/Displayed/CBPV/Unary/Instances/Displayed/FromU.agda
```diff
@@ -0,0 +1,27 @@
+-- The displayed CBPV category induced by a displayed functor Uᴰ.
+{-# OPTIONS --prop --lossy-unification #-}
```
stschaef, line 2: I don't think this `--prop` is needed either

Cubical/Categories/Displayed/CBPV/Unary/Instances/StateAlg/Base.agda
```diff
@@ -0,0 +1,493 @@
...
+open import Cubical.Categories.Displayed.CBPV.Unary.Instances.FromU
+
+private
+  variable
+    ℓ ℓ' ℓ'' ℓᴰ ℓᴰ' ℓᴰ'' ℓᴰᴰ ℓᴰᴰ' ℓD ℓD' ℓCᴰ ℓCᴰ' : Level
+
+open Functorᴰ
+
+StateAlgebra : (l : Level) → Type (ℓ-suc l)
+StateAlgebra l = Σ[ X ∈ hSet l ] StateAlg ⟨ X ⟩
```
stschaef: Displayed cat rather than a sigma?

Cubical/Categories/Displayed/CBPV/Unary/StateAlgEnrichment.agda
```diff
@@ -0,0 +1,244 @@
...
+    ℓ ℓ' ℓᴰ ℓᴰ' ℓD ℓD' ℓCᴰ ℓCᴰ' : Level
+
+open Category hiding (_∘_)
+open Categoryᴰ
+open Functor
+
+module _ (C : CBPVCat ℓ ℓ') where
+  private module C = Fibers C
+
+  StateAlgEnrichment : Type _
```
stschaef, line 53: Is there a reason this doesn't use `Categories.Monoidal.Enriched`? The interface that it provides is a little underbaked, but I think that we'd likely want to use or expand some general constructions about enrichment here

Cubical/Categories/Displayed/Functor/More.agda
```diff
@@ -160,7 +160,10 @@ module _ {C : Category ℓC ℓC'}
   where
   open Functorᴰ
   _^opFⱽ : Functorⱽ (Cᴰ ^opᴰ) (Dᴰ ^opᴰ)
-  _^opFⱽ = reindF' _ Eq.refl Eq.refl (Fⱽ ^opFᴰ)
```
stschaef, line 163: Can you explain this change?

Cubical/Categories/Displayed/Instances/Opposite.agda
```diff
@@ -0,0 +1,21 @@
+{-
+   Doubly Displayed Opposite Category.
+
+   The fact that we have to do this is a big downside of making Categoryᴰ no-eta-equality
```
stschaef, line 4: do we have any notes or issues documenting the original motivation for using `no-eta-equality`. As time goes on, I am increasingly in favor of using `eta-equality` everywhere I can, but I think this is largely a symptom of forgetting the original pain points

stschaef (commented):
There's a lot here, and I need to go back through with a finer-grained analysis: particularly on some of the doubly-displayed constructions and to step through some of the things that involved `reind`s that we would hope to elide

That's all to say, here are some initial questions/comments but I still want to take a closer look

Cubical/Algebra/State.agda
maxsnew: I'm working on this in a followup PR. I think it's fine to merge this one for now

Cubical/Categories/Displayed/CBPV/Unary/StateAlgEnrichment.agda
maxsnew: I don't even know how I would frame this in terms of enriched categories. This is just an enrichment on a profunctor, but the category overall is not enriched in algebras. So I don't think any of it is very directly applicable.

Cubical/Categories/Displayed/Instances/Opposite.agda
maxsnew: Using `eta-equality` records has *really* bad inference behavior, i.e., lots of yellow. This is one of the reasons we have so much --lossy-unification in the library, which is generally worth avoiding. I had some performance numbers as well in the original PR: https://github.com/agda/cubical/pull/1247

Cubical/Categories/Displayed/Functor/More.agda
maxsnew: This surprised me but this is actually an improvement to definitional behavior, specifically the proofs of preservation of identity and composition in reindF' currently involve transports so this makes some displayed proofs worse unfortunately. This maybe points to us needing to make a version of reindF which takes equality proofs on all of the fields of a functor, rather than just objects and morphisms

Cubical/Categories/Displayed/CBPV/Unary/Instances/Displayed/FromU.agda
maxsnew: This one is necessary because it imports things that use --prop

Cubical/Categories/Displayed/CBPV/Unary/Instances/StateAlg/Base.agda
maxsnew: Fixed this and also used EqReindex to get rid of the manual displayed stateAlgFam construction

Gluing/CBPV/StateAlg/Multiplicative.agda
```diff
@@ -0,0 +1,222 @@
...
+        (cong ([wt] (rawState raw false .fst))
+          (cong₂ seqS
+            (sym (rawRelated raw false .snd)) (sym [ret]'≡ret)))
+      ∙ raw .fst .snd
+
+  closed-FBool-surjective : ∀ M →
+    fiber interpretFreeStateBool M
+  closed-FBool-surjective M = realizeRaw (raw-FBool M)
+
+  unquote-FBool : Tm _ (gen 𝟙) ([F] (gen BoolTy)) → ⟨ FreeStateAlgebra (Bool , isSetBool) .fst ⟩
```
bond15, line 204: It is also the case that this is an [isomorphism of algebras](https://github.com/um-catlab/cubical-categorical-logic/blob/08f3f1ea123e29aca36f4ea76ff1adb0b56ed90b/HyperDoc/Examples/UF1%2Bderived.agda#L162) using this setup ?

maxsnew: Yes, that's what the commented out `unquote-quote` is trying to prove.

Cubical/Categories/Displayed/CBPV/Unary/Instances/StateAlg/Vertical.agda
```diff
@@ -0,0 +1,369 @@
...
+                  ((σ (ϕ .snd) (Bᴰ .snd) (B' .fst .snd) ⋆Homoᴰ χᴰ .snd)
+                    (Z .fst .snd))) )))
+    ∙ (Dᴰ.rectify $ Dᴰ.≡out $ Dᴰ.≡in
+      {pth = ∫Homo≡ _ _ (Z .fst .snd) refl}
+      (ΣPathP
+        ( (funExt λ b' → funExt λ x →
+            recPush-η-fᴰ (ψ .snd) Zᴰ χᴰ b' x)
+        , isProp→PathP (λ i → isPropHomoᴰ (λ z → Zᴰ .fst z .snd)) _ _ )))
+
+-- The F lift has FreeStateAlgⱽ as its intended vertex; its unit is the
```
bond15, line 312: Trying to understand how this compares to the [version I have](https://github.com/um-catlab/cubical-categorical-logic/blob/08f3f1ea123e29aca36f4ea76ff1adb0b56ed90b/HyperDoc/Operational/Effects/Instances.agda#L319) which is defined to be a type closed under the structure of the algebra.

bond15: Ah. I didn't see that you are using the state monad as the free state algebra and not some syntactic free algebra

Cubical/Categories/Displayed/CBPV/Unary/Instances/Displayed/FromProf.agda
```diff
@@ -0,0 +1,110 @@
...
+
+    Homᴰ : ∀ {k k'} {k≤ : KIND [ k , k' ]}
+      {A : Base.ob[ k ]} {B : Base.ob[ k' ]}
+      → Categoryᴰ.Hom[_][_,_] (Prof→CBPV O) k≤ A B
+      → Obᴰ A → Obᴰ B → Type ℓᴰ'
+    Homᴰ {𝓥} {𝓥} f Aᴰ Bᴰ = Vᴰ [ f ][ Aᴰ , Bᴰ ]
+    Homᴰ {𝓥} {𝓒} p Aᴰ Bᴰ = Hetᴰ p Aᴰ Bᴰ
+    Homᴰ {𝓒} {𝓒} f Aᴰ Bᴰ = Cᴰ [ f ][ Aᴰ , Bᴰ ]
+
+  -- Feels like we shouldn't have all of these rectifies...
```
bond15, line 60: Is that an artifact of using a definition of Profunctor which is a functor into a presheaf category rather than a redundant bifunctor?

maxsnew: I don't think so. I think this structure should be definitionally isomorphic to the separate action version of redundant bifunctor but I could be wrong

Cubical/Categories/Displayed/CBPV/Unary/StateAlgEnrichment.agda
```diff
@@ -0,0 +1,244 @@
...
+
+-- A CBPV model is state-enriched when its Hom profunctor extends to a
+-- profunctor valued in StateAlg. For now at least we define this
+-- manually, but it could be stated in those terms if state algebras
+-- were defined as a displayed category over sets.
+
+-- We define a displayed version in the obvious way. The most
+-- important theorem is the reindexing theorem which says that a
+-- displayed state enriched model can be reindexed along an enriched
+-- functor. For now this theorem is very ugly and manual. Maybe we can
```
bond15, line 14: Not actually an enriched functor though. Keep bumping into this distinction between "being enriched in" vs genuinely meaning "enriched". I've been using both loosely but It could be more misleading when extending beyond the unary case

Cubical/Categories/Displayed/CBPV/Unary/Instances/StateAlg/Vertical.agda
maxsnew: Yea I think this part will be similar to the free algebras where we will have a bespoke concrete version for each theory

Cubical/Categories/Displayed/CBPV/Unary/StateAlgEnrichment.agda
maxsnew: I guess we could call it "computationally enriched" but it just seems like we would always shorten this to "enriched" anyway. I don't want to get sidetracked with bikeshedding on a PR though. Start a thread on zulip if you want to discuss more

stschaef (approved):
This is good to merge

The only feedback I have I think is out of scope for now
1. being generalizations: to arbitrary theories and to arbitrary mode theories (if possible)
2. questioning if doubly displayed things are needed at all. I don't think I fully grok why we need too notions, displayed and doubly displayed. From our convo yesterday, I can by that maybe we don't need to define base/closed structure and we could instead ask for open structure that is instantiated at a closed context; this choice reduces the core of things we need to take as primitive. However, I don't fully see how/why we need to introduce another layer above displayed stuff by introducing doubly displayed stuff. Intuitively it really feels like there ought to be a single general notion, but we're still ultimately giving two

although this is maybe a moot point because it seems that the doubly displayed stuff is bootstrapped with the displayed stuff. idk I think i need to sit with it more. in any case, none of these concerns need to hold up this PR

maxsnew:
1. The definition of displayed cartesian lifts, proof that you can construct displayed cartesian lifts from cartesian lifts are all not tied to CBPV. All of the CBPV stuff is just an instantiation of abstract nonsense so it's already reusable to whatever mode theory you choose.
2. I don't know what specific doubly displayed thing you are objecting to

PR #271, opened 2026-08-11
Add --prop to Everything.agda

maxsnew:
I think that this should fix the CI failures but want to check with Steven


stschaef (approved):
Yeah this should fix it

PR #274, opened 2026-08-14
Levy local state

bond15:
Levy's Local State Monad (Thesis Chapter 7)
Includes
- Levy's local state monad as a chain of 3 adjunctions
- Proofs of some local state laws.


Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Base.agda
```diff
@@ -0,0 +1,238 @@
+module Cubical.Categories.Monad.Instances.LocalState.Levy.Equations.Base where
```
bond15, line 1: Everything up to and including this file type check in a reasonable amount of time (few seconds).
Half of the equations typecheck in seconds (example AllocGetFresh) but some take minutes (example SetGetCommute, AllocSetOld)

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/SetGetCommute.agda
```diff
@@ -0,0 +1,67 @@
...
+open Functor
+open NatTrans
+open PshHom
+
+{- A write and a read at distinct locations commute.
+
+  set i b (get j (λ c → k c))
+    = get j (λ c → set i b (k c))        when i ≢ j
+-}
+set-get-commuteᵗ : ∀ {Γ A}
```
bond15, line 23: This file takes minutes to typecheck. Nothing crazy is happening here. My one guess is that the laws are expressed as morphisms in a presheaf category and use the CC structure.

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Base.agda
```diff
@@ -0,0 +1,238 @@
...
+open import Cubical.Categories.Monad.Instances.LocalState.Levy.Discrete
+
+open Functor
+open NatTrans
+open PshHom
+
+Val-CCC : CartesianClosedCategory _ _
+Val-CCC = Cubical.Categories.Presheaf.CCC.𝓟-CCC (World ^op) ℓ-zero
+
+module V = CartesianClosedCategory Val-CCC
```
bond15, line 30: Maybe this is the issue? module unpacking?

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/SetGetCommute.agda
bond15: @stschaef @maxsnew  This is low priority but I am wondering if y'all have any insights as to why this file takes minutes to typecheck on its own. All the preliminary definitions building up the monad take seconds to check.

maxsnew: try converting any `let`s to `where`s. I was specifically warned not to use `let` for performance reasons by Jacques so that's the first thing I would check

bond15: Unfortunately this did not work. 
I'm not familiar with the exact details of the _module performance issue_
but I'm wondering if 

module V = CartesianClosedCategory (Cubical.Categories.Presheaf.CCC.𝓟-CCC (World ^op) ℓ-zero)

and usage of V.\times, V.lda, V.,p, V.\pi, etc. is leading to some kind of blowup. 
The annoying thing is that none of the definitions of the local state monad and its operations are problematic (w.r.t. time spent checking). Just the proofs of the laws.

bond15: Was able to avoid the issue with unfolding large CCC expressions using opacity.
Opacity is only used when proving equations, not in the core definitions.

Cubical/Categories/Presheaf/KanExtension/Discrete.agda
```diff
@@ -0,0 +1,155 @@
...
+private
+  variable
+    ℓI ℓD ℓD' ℓS : Level
+
+open Category
+open Functor
+open NatTrans
+open UnitCounit
+
+module _ (ℓS : Level) {I : Type ℓI} (isSetI : isSet I)
```
bond15, line 30: I think the proper way to implement this is to use the existing Presheaf/Kan.agda file and have an additional module where the functor J is assumed to be a discrete (op)fibration. This is the style of Shapiro in Familial Monads as Higher Category Theories Appendix B.

Cubical/Categories/Instances/Discrete/More.agda
```diff
@@ -0,0 +1,36 @@
+module Cubical.Categories.Instances.Discrete.More where
+
```
bond15, line 2: A definition of a discrete category which uses Eq instead of path.

Cubical/Categories/Monad/Instances/LocalState/Levy/EndCoend.agda
```diff
@@ -0,0 +1,143 @@
+open import Cubical.Foundations.Prelude
+open import Cubical.Foundations.HLevels using (hSet)
```
bond15: This could probably just be deleted.. It is the same monad but "un normalized" it uses Presheaf/Kan.agda when we can avoid the explicit end and coend.

Cubical/Categories/Presheaf/KanExtension/Discrete.agda
maxsnew: it really depends on what you want to use it for, but generally a functor with the condition of being a discrete fibration is going to be a lot worse definitionally than representing it as a Presheaf and then constructing the corresponding discrete fibration as its category of elements.

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
```diff
@@ -0,0 +1,296 @@
...
+includeOp = EqDiscFunc (λ n → n)
+
+Val : (ℓ : Level) → Category (ℓ-suc ℓ) ℓ
+Val ℓ = PresheafCategory (World ^op) ℓ
+
+Comp : (ℓ : Level) → Category (ℓ-suc ℓ) ℓ
+Comp ℓ = PresheafCategory World ℓ
+
+WorldFam : (ℓ : Level) → Category (ℓ-suc ℓ) ℓ
+WorldFam ℓ = PresheafCategory |World| ℓ
```
stschaef, line 69: I think you could get a lot of these all at once by using `Presheaf.Family.Base`, which has the presheaves/families adjunctions in it

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
```diff
@@ -0,0 +1,296 @@
...
+|World| = EqDiscreteCategory ℕ isSetℕ
+
+include : Functor |World| World
+include = EqDiscFunc (λ n → n)
+
+includeOp : Functor |World| (World ^op)
+includeOp = EqDiscFunc (λ n → n)
+
+Val : (ℓ : Level) → Category (ℓ-suc ℓ) ℓ
+Val ℓ = PresheafCategory (World ^op) ℓ
```
stschaef, line 63: In my experience, `PresheafCategory` isn't as nice to use as `PRESHEAF` from `Presheaf.StrictHom.Base` because of strictness

Although, it doesn't seem like this gets in the way of anything you're doing here; so, idk if it ought to be changed but I wanted to flag it

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
```diff
@@ -0,0 +1,296 @@
+open import Cubical.Foundations.Prelude
```
stschaef, line 1: can you put a brief descriptor of what construction this file is building? Or maybe link to a resource for it?

stschaef (commented):
a couple small comments

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
maxsnew: Yea it looks like a lot of this PR is redundant with constructions in Presheaf.Family.Base and should be removed

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Base.agda
```diff
@@ -0,0 +1,432 @@
...
+  helper-dec : (j : Fin n) (d : Dec (i ≡ j)) →
+    decRec (λ _ → σ i) (λ _ → σ j) d ≡ σ j
+  helper-dec j (yes i≡j) = cong σ i≡j
+  helper-dec j (no _) = refl
+
+  helper : (j : Fin n) → updateStore {n} i (σ i) σ j ≡ σ j
+  helper j = helper-dec j (discreteFin {n = n} i j)
+
+opaque
+  get-set-current-store : ∀ {Γ n m}
```
maxsnew, line 161: why is this equation opaque and none of the others are?

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
bond15: nice. will pull that in

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Base.agda
bond15: These opaque flags are the result of a Codex pass that was tasked with fixing performance issues. I can do a manual check that **all** opaque flags were needed to avoid blowup. The ones I tried removing caused issues. Opacity is isolated in the Equations folder.

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
bond15: Is this referring to Presheaf/KanExtension/Discrete.agda?

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
bond15: Good to know. I can try substituting that after the other edits.

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Base.agda
bond15: removed

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
```diff
@@ -0,0 +1,319 @@
+{-
+Levy's local-state monad
+
+The composing adjunctions:
+
+       Free World                -×S              PSH→Fam (World ^op)
+     ←──────────────          ←────────          ←─────────────────────
+Comp        ⊥        WorldFam     ⊥     WorldFam           ⊥            Val
```
maxsnew, line 8: Comp on the Left and Val on the right? Blasphemous

Cubical/Categories/Monad/Instances/LocalState/Levy/Base.agda
```diff
@@ -0,0 +1,319 @@
...
+T : Functor Val Val
+T = U ∘F F
+
+LS : Monad Val
+LS = T , MonadFromAdjunction F U F⊣U
+
+strength : (P A : Val .ob) →
+  Val [ P ×Psh (T .F-ob A) , T .F-ob (P ×Psh A) ]
+strength P A .N-ob n (x , t) m n≤m σ with t m n≤m σ
+... | p , m≤p , a , τ =
```
maxsnew, line 209: we don't use with abstraction in this repo

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Alloc.agda
```diff
@@ -0,0 +1,275 @@
...
+
+     alloc b (λ i → alloc c (λ j → k i j))
+       = alloc c (λ j → alloc b (λ i → k i j)).
+
+   Both sides return `suc (suc m)`, but they assign the two concrete final
+   positions in opposite orders. `World` is the preorder of natural numbers
+   and extensions; it has no permutation morphisms. Consequently there is
+   no renaming which exchanges the two fresh `Fin` positions and the matching
+   store cells, so the two computations are not equal in general.
+-}
```
maxsnew, line 275: Add explicit counter-examples to these?

Cubical/Categories/Monad/Instances/LocalState/Levy/Equations/Alloc.agda
```diff
@@ -0,0 +1,275 @@
...
+            (suc m) (γ⁺ , fresh) (suc m) ≤-refl τ)
+      ∙ cong (extendResult A ≤-sucℕ)
+          (cong (λ υ → k .N-ob (suc m) (γ⁺ , fresh)
+            (suc m) ≤-refl υ) store-path)
+      ∙ sym (allocᵗ-β {Γ = Γ} {A = A} c k n γ m n≤m σ))
+
+{- Reading the freshly allocated location returns its initial value.
+
+  alloc b (λ i → get i (λ c → k i c))
+    = alloc b (λ i → k i b)
```
maxsnew, line 67: Isn't this one derivable from alloc-set and set-get laws?

maxsnew (changes requested):
Some small things

